PR #1, opened 2019-10-14
Homework1: Anton Koliada

JustgoodDeal:
Testing pull request.


belovmd (changes requested):
1. please if u use VS code add .vscode/ line to you gitignore. It shouldn't be in PR diff
2. please add some more tasks

belovmd (approved):
8

PR #2, opened 2019-10-15
Homework1: Olga Akhmetova

Trylar:


belovmd (changes requested):
1. why do we need __init__.py file? 
2. please add some more tasks

Trylar:
Deleted empty init files since they are no longer needed since 3.3

belovmd:
9

PR #3, opened 2019-10-16
'Homework1: Pavel Martysiuk'

PavelMartysiuk:
Homework 1


belovmd (commented):
You can try some more tasks if you want. Otherwise it woul be just accepted

belovmd (commented):
7.
please create feature branch, don't use master

PR #9, opened 2019-10-17
Homework1: Andrey Kapitonov

akapitonov1999:
Homework1: Andrey Kapitonov


belovmd (changes requested):
1. please move your changes to homework branch. 
2. Something weng wrong and currently only one line is commited (`print('Hello, Maxim!')`). As I see in commits history, you've made much more tasks. It seems you need just to fix git issues.

belovmd (changes requested):
Please add more tasks

PR #10, opened 2019-10-17
Homework1: Nikolai Potovin

adnike:
Check my code. pls


belovmd (commented):
You can add some more tasks if you want. otherwise it would be just accepted later

belovmd (commented):
7

belovmd:
7

PR #14, opened 2019-10-18
Homework1: Sergey Streltsov

Mantikor:
i have moved all examples to one file, maybe that's wrong decision

### DIFF
--- a/src/homework01.py
+++ b/src/homework01.py
@@ -15,8 +15,8 @@ from itertools import groupby
 import random
 import re
 import sys
-import unittest
 from time import localtime
+import unittest
 import xml.etree.ElementTree as etree
 
 # 1 line: Output
@@ -148,6 +148,7 @@ class TestMedian(unittest.TestCase):
 # 14 lines: Doctest-based testing
 def median(pool):
     """Statistical median to demonstrate doctest.
+    
     >>> median([2, 9, 9, 7, 9, 2, 4, 5, 8])
     6 #change to 7 in order to pass the test
     """
--- a/src/homework01.py
+++ b/src/homework01.py
@@ -131,7 +131,7 @@ print(my_account.balance, my_account.overdrawn())
 
 
 # 13 lines: Unit testing with unittest
-def median(pool):
+def median_n(pool):
     copy = sorted(pool)
     size = len(copy)
     if size % 2 == 1:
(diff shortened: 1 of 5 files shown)


src/homework01.py
```diff
@@ -0,0 +1,327 @@
...
+friends = ['john', 'pat', 'gary', 'michael']
+for i, name in enumerate(friends):
+    print('iteration {iteration} is {name}'.format(iteration=i, name=name))
+
+# 4 lines: Fibonacci, tuple assignment
+parents, babies = (1, 1)
+while babies < 100:
+    print('This generation has {0} babies'.format(babies))
+    parents, babies = (babies, parents + babies)
+
```
belovmd, line 43: why previous tasks were separated with one empty line and this one with two?

src/homework01.py
```diff
@@ -0,0 +1,327 @@
...
+
+# 5 lines: Functions
+def greet(name):
+    print('Hello', name)
+
+
```
belovmd, line 49: why this function's calls are separated from this function with 2 lines and next task with one from calls?

src/homework01.py
```diff
@@ -0,0 +1,327 @@
...
+%d bottles of beer on the wall,
+%d bottles of beer,
+take one down, pass it around,
+%d bottles of beer on the wall!
+'''
+bottles_of_beer = 9
+while bottles_of_beer > 1:
+    print(REFRAIN % (bottles_of_beer, bottles_of_beer, bottles_of_beer - 1))
+    bottles_of_beer -= 1
+
```
belovmd, line 120: why here we have 2 lines between tasks and in previous one?

src/homework01.py
```diff
@@ -0,0 +1,327 @@
...
+import random
+import re
+import sys
+from time import localtime
+import unittest
+import xml.etree.ElementTree as etree
+
+# 1 line: Output
+print('Hello, world!')
+
```
belovmd, line 25: it looks as a good idea to separate tasks from each other with 2 empty lines everywhere

src/homework01.py
```diff
@@ -0,0 +1,327 @@
...
+This is the second.
+'''.splitlines()
+# Use itertools.groupby and bool to return groups of
+# consecutive lines that either have content or don't.
+for has_chars, frags in groupby(lines, bool):
+    if has_chars:
+        print(' '.join(frags))
+
+
+# PRINTS:
```
belovmd: do we need this comments here?

belovmd (changes requested):
This code is OK itself but looks pretty weird. Please address issues below
and please fix issues like that everywhere

src/homework01.py
Mantikor: this recommendation comes from PyCharm, PEP8: expected 2 blank lines, found 1, i think this problem is due i join all tasks in one file

belovmd (commented):
8:
1.  unnecessary files (from 2nd homework)
2. used master branch (in guidelines were suggested to use feature branch)

PR #17, opened 2019-10-19
Homework1: Aliaksandr Khalupka

Alex27Khalupka:


src/2.py
```diff
@@ -0,0 +1,40 @@
...
+parents, babies = (1, 1)
+while babies < 100:
+    print('This generation has {0} babies'.format(babies))
+    parents, babies = (babies, parents + babies)
+
+
+def greet(par):
+    print('Hello', par)
+
+
```
belovmd: it looks strange that our greet function separated from it's calls with 2 empty lines and next task with re with only one

belovmd (commented):
1. please take a look at the issue below
2. If you want you can add more tasks.  otherwise it can be accepted as it is

src/2.py
```diff
@@ -0,0 +1,40 @@
...
+    else:
+        print(test_string, 'rejected')
+
+prices = {'apple': 0.40, 'banana': 50}
+my_purchase = {
+    'apple': 1,
+    'banana': 6
+}
+grocery_bill = sum(prices[fruit] * my_purchase[fruit] for fruit in my_purchase)
+# print(my_purchase[fruit] for fruit in my_purchase)
```
belovmd: do we need this comment?

belovmd (changes requested):
1. Please put task description to docstring or comment
2. Why first one is separated? it should be either each in it's file or all in one (overall I prefer first approach)
3. You can add some more tasks if you want

belovmd (commented):
9
1. Please add tasks description next time
2. please don't delete __init__.py

PR #19, opened 2019-10-19
Homework2: Alex Levdansky

Coromilly:


Homework1.py
```diff
@@ -0,0 +1,324 @@
...
+import random
+import re
+import sys
+from time import localtime
+import unittest
+import xml.etree.ElementTree as Etree
+
+# 1
+print('Hello, world!')
+# 2
```
belovmd, line 16: It would be much more readable if we separate tasks with two empty lines

Homework1.py
```diff
@@ -0,0 +1,324 @@
...
+parents, babies = (1, 1)
+while babies < 100:
+    print('This generation has {0} babies'.format(babies))
+    parents, babies = (babies, parents + babies)
+
+
+# 5
+def greet(nane):
+    print('Hello', nane)
+
```
belovmd, line 37: no sense to separate with two lines function and it's calls in such a big file

Homework1.py
```diff
@@ -0,0 +1,324 @@
...
+my_purchase = {
+    'apple': 1,
+    'banana': 6}
+grocery_bill = sum(prices[fruit] * my_purchase[fruit]
+                   for fruit in my_purchase)
+print('I owe the grocer $%.2f' % grocery_bill)
+# 8
+# This program adds up integers that have been passed
+# as arguments in the command line
+
```
belovmd, line 65: it looks strange our docstring goes after previous without empty line and separated from its solution.

belovmd (changes requested):
Please reformat this code a bit to make it a bit more readable

fizz_buzz.py
```diff
@@ -0,0 +1,9 @@
+for number in range(101):
+    if number % 3 == 0 and number % 5 == 0:
```
belovmd, line 2: you can use 15 here instead

money_calculation.py
```diff
@@ -0,0 +1,12 @@
+price = str(input('Input price, separate dollars from cents with "." : '))
+num_of_items = int(input('Input number of items you want to buy: '))
+dollars = int(price.split('.')[0])
+cents = int(price.split('.')[1])
+print('One item price is {0} dollars {1} cents'.format(dollars, cents))
```
belovmd, line 5: In description wasn't mentioned this print

money_calculation.py
```diff
@@ -0,0 +1,12 @@
...
+if cents * num_of_items >= 100:
```
belovmd, line 6: we can avoid this if condition here

money_calculation.py
```diff
@@ -0,0 +1,12 @@
+price = str(input('Input price, separate dollars from cents with "." : '))
```
belovmd, line 1: I will enter 12.23423141234

task3.py
```diff
@@ -0,0 +1,10 @@
+def checkio(number: int) -> int:
+    """Calculate the product of the digits excluding any zeroes."""
```
belovmd, line 2: Please put more clear description here

task5.py
```diff
@@ -0,0 +1,10 @@
+def checkio(array):
+    """sums even-indexes elements and multiply at the last."""
+    sum = 0
+    if array == []:
```
belovmd, line 4: `if not array` would be enough?

belovmd (changes requested):
1. please remove files from HW1
2. Please solve Palindrome task without strings methods - only using arithmetical operations
3. Please address these issues:

task5.py
Coromilly: Solved.

task3.py
Coromilly: fixed.

fizz_buzz.py
Coromilly: fixed.

money_calculation.py
Coromilly: Fixed

money_calculation.py
Coromilly: solved.

money_calculation.py
Coromilly: solved.

PR #36, opened 2019-10-23
Homework2: Sergey Streltsov

Mantikor:
seems every new homework need to put in new folder for creating separate pull request to each homework?


src/homework02_fizzbuzz.py
```diff
@@ -0,0 +1,20 @@
...
+
+"""
+EPAM python q4 homework 02.3 FizzBuzz
+Sergey Streltsov 2019-10-22
+"""
+
+
+if __name__ == '__main__':
+    for number in range(1, 101):
+        if (number % 3 == 0) and (number % 5 == 0):
```
belovmd: you can use 15 instead

src/homework02_palindrome_v2.py
```diff
@@ -0,0 +1,25 @@
...
+    return num != 0 and ((num % 10) * (10 ** int(math.log(num, 10))) +
+                         reverse_number(num // 10))
+
+
+if __name__ == '__main__':
+    input_str = input('Enter a number: ')
+    try:
+        num_value = int(input_str)
+        print('Is palindrome: {}'.format(
+            num_value == reverse_number(num_value)))
```
belovmd: it could be easier solved easier without recursive functions with one for loop
it looks implicit and too complex.

belovmd (changes requested):
1. please remove 1st HW files from pr
2. please put tasks descriptions to files. 
3. please address these issues:

belovmd:
> seems every new homework need to put in new folder for creating separate pull request to each homework?

It would be enough to use separate branch for each HW

belovmd (changes requested):
I don't see homework 2 files

PR #37, opened 2019-10-23
Homework2:Pavel Martysiuk

PavelMartysiuk:


src/task3.py
```diff
@@ -0,0 +1,16 @@
+def fizz_bazz(range_):
+    for digit in range(range_):
+        if digit % 3 == 0 and digit % 5 == 0:
```
belovmd: you can use 15 here

src/task1.py
```diff
@@ -0,0 +1,6 @@
+def money_calculator(cost, quantity):
```
belovmd: please solve it according to task description. 
input: M, N, L etc

belovmd (changes requested):
1. please remove tasks from first HW
2. please put tasks descriptions to docstrings
3. please add palindrom task
4. please address next issues:

src/first_word.py
```diff
@@ -0,0 +1,15 @@
+'''Print first word in string'''
+
+
+def first_word(str_):
+    word = ''
+    for sym in str_:
+        if sym == ' ':
```
belovmd, line 7: it wouldn't work with `Hello, world'

src/isogram.py
```diff
@@ -0,0 +1,18 @@
+''' Write program that check str is isogram.'''
```
belovmd, line 1: no description what is isogram in this tasks (there are some options for in)

src/isogram.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def iso(str_):
+    str_ = str_.lower()
+    dicty = {}
+    for sym in str_:
```
belovmd, line 7: it's complexity is O(n**2) - too complex (count inside for)

src/isogram_2.py
```diff
@@ -0,0 +1,12 @@
+'''Isogram sedond version'''
+
+
+def iso(str_):
+    if len(str_) == len(set(str_.lower())):
```
belovmd, line 5: `return len(str_) == len(set(str_.lower()))` would be enough

src/max_cost.py
```diff
@@ -0,0 +1,9 @@
+'''Find max value in dict'''
+
+
+def max_cost(dict_):
+    return max(dict_.items(), key=lambda x: x[-1])
```
belovmd, line 5: do we really need this lambda?

src/power.py
```diff
@@ -0,0 +1,10 @@
+'''raise to the power args'''
```
belovmd, line 1: 1. Unclear description
2. please use tripple double quotes

src/task1.py
```diff
@@ -0,0 +1,12 @@
+'''Purchase money calculator'''
```
belovmd, line 1: """ please put docstring """
unspecified input

src/task2.py
```diff
@@ -0,0 +1,18 @@
+'''Palindrome'''
```
belovmd, line 1: """

src/task2.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def palindrome(num):
```
belovmd, line 4: unspecified input

src/task2.py
```diff
@@ -0,0 +1,18 @@
...
+    power = len(str(num)) - 1
+    for i in range(1, (len(str(num)) // 2) + 1):
+        first_digit = num // 10 ** power
```
belovmd, line 7: if we just revert num it would be much more clear

src/task3.py
```diff
@@ -0,0 +1,19 @@
+'''FizzBazz'''
```
belovmd, line 1: """FixxBUzz"""

src/task3.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def fizz_bazz(range_):
+    for digit in range(range_):
```
belovmd, line 5: it would pring 0-99 instead of 100

belovmd (commented):
1. why do you need to move \_\_init__ ?

7.5

src/max_cost.py
PavelMartysiuk: Yes, if  I don't use lambda program return a max key.

PavelMartysiuk: Yes, if I don't use the lambda function, the function return a max key of dict.

PR #40, opened 2019-10-25
Homework1: Maksim Bleo

Meolax:


belovmd (commented):
it scores as 7.  Task looks good(aprox. 9) but there are some issues that make PR weird:
1. Added not needed files for docker
2. Added not needed vscode settings file
3. removed \_\_init__.py (why?)

PR #41, opened 2019-10-25
Homework2: Aliaksandr Khalupka

Alex27Khalupka:


Homework2/6.py
```diff
@@ -0,0 +1,25 @@
+"""
+    sums even-indexes elements and multiply at the last
+"""
+
+
+def checkio(array):
+
+    if len(array) == 0:
```
belovmd, line 8: would it be enough `if array`?

Homework2/6.py
```diff
@@ -0,0 +1,25 @@
...
+        return 0
+
+    sum = 0
+    element = 0
```
belovmd, line 12: do we need this assignment?

Homework2/Task1.py
```diff
@@ -0,0 +1,16 @@
+"""
+1-st task from Homework1
+"""
+
+str1 = input("Enter your string: ")
+sim = len(str1) // 2 + len(str1) % 2
+i = 0
+
+while i < sim:
```
belovmd, line 9: please use numbers and arithmetic operators

Homework2/Task3.py
```diff
@@ -0,0 +1,13 @@
+"""
+3-rd Task from Homework2
+"""
+
+for number in range(1, 101):
+    if (number % 3 == 0) and (number % 5 == 0):
```
belovmd, line 6: can we use 15 here?

Homework2/Task5.py
```diff
@@ -0,0 +1,34 @@
...
+Task from Checkio.
+returns the first word in a given text.
+"""
+
+
+def first_word(text: str) -> str:
+    text = text + " "
+    word = ""
+    f = False
+    not_letters = " !.,?:\\/"
```
belovmd, line 11: can we use split and strip here instead?

belovmd (changes requested):
1. please remove tasks from HW1
2. please don't remove \_\_init__.py file
3. Please address these issues:

belovmd (commented):
6.5

PR #43, opened 2019-10-26
Homework1: Alena Antipina

windsdel:
Changing style for the programs from https://wiki.python.org/moin/SimplePrograms

### DIFF
--- a/HW1_1.py
+++ b/HW1_1.py
@@ -1,0 +1,2 @@
+print ('Hello, world!')
+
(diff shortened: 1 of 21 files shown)


belovmd (commented):
I'll review it tomorrow to give you opportunity to fix codestyle issues as I see you are doing it right now

belovmd (changes requested):
please fix code style

PR #44, opened 2019-10-26
Homework2 Nikolai Potovin

adnike:
5 examples from https://py.checkio.org/  on the top docstring and 3 examples below

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,0 +1,4 @@
+parents, babies = (1, 1)
+while babies < 100:
+    print ('This generation has {0} babies'.format(babies))
+    parents, babies = (babies, parents + babies)
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,10 +1,11 @@
 parents, babies = (1, 1)
 while babies < 100:
-    print ('This generation has {0} babies'.format(babies))
+    print('This generation has {0} babies'.format(babies))
     parents, babies = (babies, parents + babies)
 
+
 def greet(name):
-    print ('Hello', name)
+    print('Hello', name)
 
 greet('Jack')
 greet('Jill')
@@ -16,4 +17,4 @@ my_purchase = {
     'banana': 6}
 grocery_bill = sum(prices[fruit] * my_purchase[fruit]
                    for fruit in my_purchase)
-print ('I owe the grocer $%.2f' % grocery_bill)
+print('I owe the grocer $%.2f' % grocery_bill)
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -8,19 +8,19 @@
 
 
 # PALINDROME
-# print('input number')
-# n = int(input())
-# ONum = n
-# ost = 0
-# RNum = 0
-# while n != 0:
-#     ost = n % 10
-#     RNum = RNum * 10 + ost
-#     n = n // 10
-# if ONum == RNum:
-#     print(ONum, 'is a palindrome')
-# else:
-#     print(ONum, 'is not a palindrome')
+print('input number')
+n = int(input())
+ONum = n
+ost = 0
+RNum = 0
+while n != 0:
+    ost = n % 10
+    RNum = RNum * 10 + ost
+    n = n // 10
+if ONum == RNum:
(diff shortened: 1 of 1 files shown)


src/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+# if x[N] doesn't exists then return (-1)
+# Digits Manipulation:Multiply all but 0 digits of positive integers.
+# Secret Message:Gather all capital letters in one word
+# in the order that they appear in the chunk of text.
+
+
+# PALINDROME
+print('input number')
+n = int(input())
+ONum = n
```
belovmd: please name variables in snake case

src/__init__.py
```diff
@@ -0,0 +1,47 @@
+# Multiply: multiply 2 numbers
+# Easy Unpack: return 1st, 3rd and second to last element from tuple
+# Index Power:power Nth element of list in the N degree,
```
belovmd: I don't understand this docstring

src/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+if ONum == RNum:
+    print(ONum, 'is a palindrome')
+else:
+    print(ONum, 'is not a palindrome')
+
+# MULTIPLIER
+print('input one item price is')
+price = float(input())
+print('input count')
+cnt = float(input())
```
belovmd: why don't we use int?

src/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    ost = n % 10
+    RNum = RNum * 10 + ost
+    n = n // 10
+if ONum == RNum:
+    print(ONum, 'is a palindrome')
+else:
+    print(ONum, 'is not a palindrome')
+
+# MULTIPLIER
+print('input one item price is')
```
belovmd: i will enter price 1.2335232435345 and it seems to be impossible cents count

src/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+cnt = float(input())
+summa = 0
+summa = (price * cnt)
+print('your total=', int(summa), 'dollars',
+      round((summa - int(summa)) * 100), 'cents')
+
+# FIZZBUZZ
+num = 1
+while num <= 100:
+    if (num % 3) == 0:
```
belovmd: can't we use 15 here?

belovmd (changes requested):
1. please add tasks suggested sites. 
2. please address these issues:

src/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+# if x[N] doesn't exists then return (-1)
+# 4.Digits Manipulation:Multiply all but 0 digits of positive integers.
+# 5.Secret Message:Gather all capital letters in one word
+# in the order that they appear in the chunk of text.
+
+
+# PALINDROME
+print('input number')
+n = int(input())
+o_num = n
```
belovmd: unclear vars naming
n, o_num, r_num etc

belovmd (changes requested):
please add tasks from suggested sites

src/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+# in the order that they appear in the chunk of text.
+
+
+# PALINDROME
+print('input number')
+number = int(input())
+original_num = number
+ostatok = 0
+reversd_num = 0
+while number != 0:
```
belovmd, line 17: while num?

belovmd (commented):
5
overdue, missed external sites tasks

src/__init__.py
adnike: Maxim,
I would like to discuss some issues regarding HW2. First of all, your comment about adding task from suggested sites was sent after the deadline of 01.11.2019. So, I considered my hw being OK and checking my email for the rest of the time optional. That's why I didn't see the comment mentioned and didn't answer it. Second, I've done all the five tasks needed (see screenshot). And third, it would be very interesting for me to discuss the views on naming variables because I consider it inconvenient to name variables as full russian words since we work in the community of people who know English language and professional programmer lingo.
Best regards,
Nikolai

proofhw2

PR #46, opened 2019-10-27
Homework2: Anton Stychnevsky

Stychnevsky:
Homework2
In task 4 I use https://www.codewars.com/

### DIFF
--- a/HW2-kata1.py
+++ b/HW2-kata1.py
@@ -10,8 +10,8 @@ move_zeros([false,1,0,1,2,0,1,3,"a"]) # returns[false,1,1,2,1,3,"a",0,0]
 
 def move_zeros(array):
     array_without_zero = [i for i in array if i or (type(i) == bool)]
-    new_array = (array_without_zero
-                 + [0] * (len(array) - len(array_without_zero)))
+    new_array = (array_without_zero +
+                 [0] * (len(array) - len(array_without_zero)))
     return new_array
 
 
--- a/HW2-kata1.py
+++ b/HW2-kata1.py
@@ -1,0 +1,16 @@
+""" 5 kyu - Moving Zeros To The End https://www.codewars.com/kata/52597aa56021e91c93000cb0
+
+Write an algorithm that takes an array and moves all of the zeros to the end,
+preserving the order of the other elements.
+
+move_zeros([false,1,0,1,2,0,1,3,"a"]) # returns[false,1,1,2,1,3,"a",0,0]
+"""
+
+
+def move_zeros(array):
+    array_without_zero = [i for i in array if (i != 0) or (type(i) == bool)]
+    new_array = array_without_zero + [0]*(len(array) - len(array_without_zero))
+    return new_array
+
+
+print(move_zeros([False, 1, 0, 1, 2, 0, 1, 3, "a"]))  # for example
--- a/HW2-kata1.py
+++ b/HW2-kata1.py
@@ -9,7 +9,7 @@ move_zeros([false,1,0,1,2,0,1,3,"a"]) # returns[false,1,1,2,1,3,"a",0,0]
 
 
 def move_zeros(array):
-    array_without_zero = [i for i in array if (i != 0) or (type(i) == bool)]
+    array_without_zero = [i for i in array if i or (type(i) == bool)]
     new_array = (array_without_zero
                  + [0] * (len(array) - len(array_without_zero)))
     return new_array
--- a/HW2-kata1.py
+++ b/HW2-kata1.py
@@ -1,4 +1,5 @@
-""" 5 kyu - Moving Zeros To The End https://www.codewars.com/kata/52597aa56021e91c93000cb0
+""" 5 kyu - Moving Zeros To The End
+https://www.codewars.com/kata/52597aa56021e91c93000cb0
 
 Write an algorithm that takes an array and moves all of the zeros to the end,
 preserving the order of the other elements.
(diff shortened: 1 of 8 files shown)


Stychnevsky:
Test flake8 failed by:
W503 line break before binary operator

But, as I understand, this rule was changed on April 2016 and it is better to break before a binary operator.:
https://github.com/python/peps/commit/c59c4376ad233a62ca4b3a6060c81368bd21e85b
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

So, I didn't change my code

belovmd:
A you writing like this: 
> x = y+z
> +123

in a notebook when you are solving math equations? 

I agree that it's controversial question, but however linters should pass

HW2-kata1.py
```diff
@@ -0,0 +1,18 @@
...
+
+Write an algorithm that takes an array and moves all of the zeros to the end,
+preserving the order of the other elements.
+
+move_zeros([false,1,0,1,2,0,1,3,"a"]) # returns[false,1,1,2,1,3,"a",0,0]
+"""
+
+
+def move_zeros(array):
+    array_without_zero = [i for i in array if (i != 0) or (type(i) == bool)]
```
belovmd: `if i or (type......`

HW2-kata2.py
```diff
@@ -0,0 +1,35 @@
...
+longest_slide_down) that takes a pyramid representation as argument and
+returns its' longest 'slide down'. For example, longestSlideDown([[3], [7,
+4], [2, 4, 6], [8, 5, 9, 3]]) => 23
+
+By the way... My tests include some extraordinarily high pyramides so as you
+can guess, brute-force method is a bad idea unless you have a few centuries
+to waste. You must come up with something more clever than that. """
+
+
+def longest_slide_down(pyramid):
```
belovmd, line 27: 💯

HW2-kata3.py
```diff
@@ -0,0 +1,56 @@
...
+    previous_number = arr[0]
+    previous_pos = 0
+    answer = {"pos": [], "peaks": []}
+
+    for pos, number in enumerate(arr[1:], 1):
+        if number > previous_number:
+            previous_number = number
+            previous_pos = pos
+            direction = "uphill"
+        elif number == previous_number:
```
belovmd: `for pos, number: `
`if previous< number and number >= next: peak`
wouldn't be a bit shorter?

HW2-kata4.py
```diff
@@ -0,0 +1,87 @@
...
+mauris,  at  elementum  ligula
+tempor  eget.  In quis rhoncus
+nunc,  at  aliquet orci. Fusce
+at   dolor   sit   amet  felis
+suscipit   tristique.   Nam  a
+imperdiet tellus.
+"""
+
+
+def justify(text, width):
```
belovmd, line 36: maybe it has sense to get a list of all fords in appropriate order `words = test.split()`
next you can take from it words that no more than width (with at least one space) and write function to form string for these words? 
it seems to me it would be shorter and clearer

HW2-task3.py
```diff
@@ -0,0 +1,16 @@
+""" FizzBuzz
+Write a program that prints the numbers from 1 to 100, but for multiples of
+three print “Fizz” instead of the number and for multiples of five print
+“Buzz”. For numbers which are multiples of both three and five,
+print “FizzBuzz”.
+"""
+
+for number in range(1, 101):
+    if (number % 3 == 0) & (number % 5 == 0):
```
belovmd: 15?

HW2-task1.py
```diff
@@ -0,0 +1,24 @@
+""" Finding palindrome
+Write a program that check whether a number is palindrome or Not.
+Input string contains only numbers.
+Please work only arithmetic operations, loops and if-condition.
+"""
+
+number = int(input("Enter number: \n"))
+if number:
+    len_of_number = 0
```
belovmd: maybe it would be easier to revert a number with just one for loop? 
this solution looks too complex

HW2-kata5.py
```diff
@@ -0,0 +1,89 @@
...
+
+N.B. to keep it simplest, the string in input is restricted to represent
+only multilinear non-constant polynomials, so you won't find something like
+`-3+yx^2'. Multilinear means in this context: of degree 1 on each variable.
+
+Warning: the string in input can contain arbitrary variables represented by
+lowercase characters in the english alphabet. """
+
+
+def simplify(poly):
```
belovmd, line 35: it looks too complex but currently i have now idea how to simplify it.

belovmd (changes requested):
You homework looks great! 
Please take a look at these issues:

HW2-kata3.py
Stychnevsky: It's good idea, but there is a problem with "plateaus."
For example, [1,2,2,2,2] and [1,2,2,2,3] have no peaks.
I change code using your advice, but it needs one more loop. In spite of this
 new code looks more simply. But I think it requires more operations

HW2-kata5.py
Stychnevsky: I tried to use str.split(), but it became hard to understand how many words will be in every line, and code still looked not good

Than I looked through decisions on Codewars and found a better way to split text by lines. I used this idea, and than split words into lines.
So, I think now code became simpler.

HW2-kata3.py
belovmd: It's a valid. One more idea: find all local maximums as I suggested. 
Than we can find out all local minimums  previous>= num >=next
After that we can check that after each local max would be local min lesser than it - than it would be correct peak. Othervise (max = next min) it would be just plateau

belovmd (approved):
10

PR #50, opened 2019-10-27
Homework2: Andrey Kapitonov

akapitonov1999:


FizzBuzz.py
```diff
@@ -0,0 +1,16 @@
+# FizzBuzz task
+
+i = 0
+while i != 100:
+    i += 1
+    if i % 3 == 0 and i % 5 == 0:
```
belovmd, line 6: 15?

Codewars1.py
```diff
@@ -0,0 +1,67 @@
+''' In this kata you will create a function
+that takes a list of non-negative integers
+and strings and returns a new list with the
+strings filtered out.'''
+
+
+def filter_list(l):
+    i = 0
```
belovmd, line 8: please don't use magic names l, i

Codewars1.py
```diff
@@ -0,0 +1,67 @@
...
+    while i != len(l):
+        if type(l[i]) != int:
+            l.pop(i)
+        else:
+            i += 1
+            continue
```
belovmd, line 14: we don't need continue here

Codewars1.py
```diff
@@ -0,0 +1,67 @@
...
+    return l
+
+
+'''An isogram is a word that has no repeating letters,
+consecutive or non-consecutive. Implement a function
+that determines whether a string that contains only letters is an isogram.
+Assume the empty string is an isogram. Ignore letter case.'''
+
+
+def is_isogram(string):
```
belovmd, line 24: can we just use set here?

Codewars1.py
```diff
@@ -0,0 +1,67 @@
...
+to hear non-native Italian speakers trying to pronounce it :('''
+
+
+def tribonacci(signature, n):
+    if n == 0:
+        return []
+    elif n == 1:
+        listing = []
+        listing.append(signature[0])
+        return listing
```
belovmd, line 55: `return [signature[0]]`?

Codewars1.py
```diff
@@ -0,0 +1,67 @@
...
+    elif n == 2:
+        listing = []
+        listing.append(signature[0])
+        listing.append(signature[1])
+        return listing
```
belovmd, line 60: same as above

Palindrome.py
```diff
@@ -0,0 +1,45 @@
...
+        print('YES, it is an int palindrome!')
+    else:
+        print('NO, it is not a palindrome!')
+
+
+def strpalindrome(a):
+    x = ''
+    for i in a:
+        x = i + x
+        if a == x:
```
belovmd, line 29: just if a==a[::-1]

Palindrome.py
```diff
@@ -0,0 +1,45 @@
...
+
+def intpalindrome(a):
+    length = int(math.log10(a)) + 1
+    i = 0
+    listing1 = []
+    while i != length:
+        x = a // (10 ** i) % 10
+        listing1.append(x)
+        i += 1
+    listing2 = listing1[::-1]
```
belovmd, line 18: we don't need lists here. just build reversed number using arithmetic operator

hello.py
```diff
@@ -0,0 +1 @@
+print('Hello, Maxim!')
```
belovmd, line 1: Hi!

belovmd (changes requested):
Thank you for your work. 
Please address these issues:

Codewars1.py
akapitonov1999: l is for 'list', i is for iteration

Codewars1.py
akapitonov1999: ok

FizzBuzz.py
akapitonov1999: like if it was 'if i % 15 == 0:' ?

Codewars1.py
belovmd: better to use names that show what do they keep

belovmd (commented):
7

PR #51, opened 2019-10-27
Homework2: Pavel Ralko

par216:


1.py
```diff
@@ -0,0 +1,19 @@
...
+
+if n < 0:
+    print("NO")
+else:
+    k = n
+    m = 0
+    while k > 0:
+        m = 10 * m + k % 10
+        k = k // 10
+        print(m)
```
belovmd: redundant print

1.py
```diff
@@ -0,0 +1,19 @@
...
+
+if (m == n):
```
belovmd: if I'll enter -10 it would crash here

2.py
```diff
@@ -0,0 +1,58 @@
...
+'''
+second way
+'''
+price1_dol = 3
+price1_cen = 20
+
+cost4 = 0
+cost5 = 0
+
+for i in range(num1):
```
belovmd: can you please just make 3 inputs: M, L, N

3.py
```diff
@@ -0,0 +1,15 @@
+'''
+FizzBuzz
+'''
+for i in range(101):
+
+    if i % 5 == 0 and i % 3 == 0:
```
belovmd: 15?

belovmd (commented):
1. please add tasks from suggested sites
2. Please address these issues

2.py
```diff
@@ -0,0 +1,18 @@
+'''
+ Purchase money calculation
+'''
+
+M = 3
+N = 20
```
belovmd, line 6: missed m, n inputs

belovmd (commented):
6.5

PR #52, opened 2019-10-27
Homework2: Andrey Kuksik

Kuxsoger:


finding_palindrome.py
```diff
@@ -0,0 +1,22 @@
...
+Input string contains only numbers.
+Please work only arithmetic operations, loops and if-condition
+"""
+
+number = input()
+len_number = len(number)
+number = int(number)
+
+palindrome = True
+for left in range(len_number - 1, 0, -2):
```
belovmd: it would look like more simple if we just revert number using arithmetic operators

fizz_buzz.py
```diff
@@ -0,0 +1,17 @@
+"""
+Write a program that prints the numbers from 1 to 100,
+but for multiples of three print “Fizz” instead of the number
+and for multiples of five print “Buzz”.
+For numbers which are multiples of both three and five, print “FizzBuzz”.
+"""
+
+for num in range(1, 101):
+    if not num % 3:
```
belovmd: 15?

belovmd (commented):
Please address these minor issues

purchase_money_calculation.py
```diff
@@ -0,0 +1,19 @@
...
+Print total price in dollars and cents for L items.
+For example
+Input data:  one item price is 3 dollars 20 cents, calculate price for 3 items
+Output data: Total cost:  9 dollars 20 cents.
+"""
+
+# There were no additional explanations about input
+# so I did it exactly like in example
+input_data = input().split(' ')
+M, N, L = (int(word) for word in input_data if word.isnumeric())
```
belovmd, line 13: what would happen if it wouldn't be a number?

task4_5.py
```diff
@@ -0,0 +1,37 @@
...
+2 5 4 3 1
+
+output
+['4', 'Gold Medal', 'Silver Medal', 'Bronze Medal', '5']
+"""
+
+
+def find_relative_ranks(nums):
+    ranks = {}
+    sorted_nums = sorted(nums, reverse=True)
```
belovmd, line 20: it can be done with O(n) instead of O(n log n)

belovmd (commented):
8

PR #53, opened 2019-10-28
Homework1: Alena Antipina

windsdel:
Changing style for the programs from https://wiki.python.org/moin/SimplePrograms


belovmd (commented):
5

PR #54, opened 2019-10-29
Homework2: Anastasiya Infarovich

AnInf:
Write simple programs

### DIFF
--- a/src/task2/calculation.py
+++ b/src/task2/calculation.py
@@ -2,20 +2,19 @@
 
 One item costs M dollars and N cents. Customer bought L items.
 Print total price in dollars and cents for L items.
-
-For example
 Input data:  one item price is 3 dollars 20 cents, calculate price for 3 items
 Output data: Total cost:  9 dollars 60 cents.
 """
+
+
 dollars, cents = \
     input('Enter one item price in format dollars.cents, like 3.14: ')\
     .split('.')
 dollars = int(dollars)
 cents = int(cents)
 total_cents = dollars * 100 + cents
-# print(f"to single item price: {C} cents")
 total_items = int(input("How many items are needed? "))
 total_cost = int(total_cents * total_items)
 final_dollars = int(total_cost // 100)
 final_cents = int(total_cost % 100)
-print(f"Total cost: {final_dollars} dollars {final_cents} cents.")
+print("Total cost: {} dollars {} cents.".format(final_dollars, final_cents))
--- a/src/task2/calculation.py
+++ b/src/task2/calculation.py
@@ -1,0 +1,21 @@
+"""Calculate total cost
+
+One item costs M dollars and N cents. Customer bought L items.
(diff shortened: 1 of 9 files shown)


src/task2/say_hi.py
```diff
@@ -0,0 +1,8 @@
+"""Introduce a person with the given parameter's attributes."""
+
+
+def say_hi(name: str, age: int) -> str:
+    return "Hi. My name is {} and I'm {} years old".format('Alex', 32)
```
belovmd, line 5: `.format(name, age)`
I suppose

src/task2/palindrome.py
```diff
@@ -0,0 +1,19 @@
...
+Check whether a number is palindrome or Not.
+Input string contains only numbers.
+"""
+
+
+number_to_check = input("Enter a number: ")
+reversed_number = ""
+number_to_check_len = len(number_to_check)
+
+while number_to_check_len:
```
belovmd, line 12: work with arithmetic operators and numbers should be used according to task description

src/task2/fizz_buzz.py
```diff
@@ -0,0 +1,18 @@
...
+
+A program that prints the numbers from 1 to 100,
+for multiples of three print “Fizz” instead of the number,
+for multiples of five print “Buzz”,
+for numbers which are multiples of both three and five, print “FizzBuzz”.
+"""
+
+
+for number in range(1, 100):
+    if number % 3 == 0:
```
belovmd, line 11: 15?

src/task2/digits_multiplication.py
```diff
@@ -0,0 +1,13 @@
+"""Calculate the product of the digits excluding any zeroes."""
+
+
+def checkio(number: int) -> int:
```
belovmd, line 4: no need to use strings here. 
It would be enough to use arithmetical operators

src/task2/calculation.py
```diff
@@ -0,0 +1,20 @@
+"""Calculate total cost
+
+One item costs M dollars and N cents. Customer bought L items.
+Print total price in dollars and cents for L items.
+Input data:  one item price is 3 dollars 20 cents, calculate price for 3 items
+Output data: Total cost:  9 dollars 60 cents.
+"""
+
+
+dollars, cents = \
```
belovmd, line 10: line transfer looks weird here

src/task2/calculation.py
```diff
@@ -0,0 +1,20 @@
...
+    input('Enter one item price in format dollars.cents, like 3.14: ')\
```
belovmd, line 11: I will enter 3.111111111

belovmd (commented):
1. No need to delete \_\_init__ here - it's required
2. Overall I like tasks descriptions and code, but in some tasks it's overhead (strings instead of numbers)
7.5

PR #59, opened 2019-11-02
Homework3: Aleksandr Levdansky

Coromilly:
Third homework. Includes 3 tasks:
1) List practice;
2) Tuple practice;
3) Dictionary practice.

### DIFF
--- a/dictionary_practice.py
+++ b/dictionary_practice.py
@@ -1,6 +1,5 @@
 def generate_numbers(n=20):
     """Function returns dictionary.
-
     The keys are numbers between 1 and n (both included) and the values
     are square of keys. n – function argument. Default is 20.
     """
--- a/dictionary_practice.py
+++ b/dictionary_practice.py
@@ -8,10 +8,7 @@ def generate_numbers(n=20):
 
 
 def count_characters(s):
-    """Function count and return the numbers of each character
-    
-    in a string s.
-    """
+    """Function count and return the numbers of each character in a string s."""
     dct = {}
     for symb in s:
         if symb in dct:
--- a/dictionary_practice.py
+++ b/dictionary_practice.py
@@ -1,0 +1,20 @@
+def generate_numbers(n=20):
+    """Function returns dictionary.
+
+    The keys are numbers between 1 and n (both included) and the values
+    are square of keys. n – function argument. Default is 20.
+    """
+    return {num: num**2 for num in range(1, n + 1)}
+
+
+def count_characters(s):
+    """
+    Function count and return the numbers of each character in a string s.
+    """
+    dct = {}
+    for symb in s:
+        if symb in dct:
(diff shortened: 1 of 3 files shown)


dictionary_practice.py
```diff
@@ -0,0 +1,18 @@
...
+
+def count_characters(s):
+    """Function count and return the numbers of each character in a string."""
+    dct = {}
+    for symb in s:
+        if symb in dct:
+            dct[symb] += 1
+        else:
+            dct[symb] = 1
+    return {k: dct[k] for k in sorted(dct)}
```
belovmd: do we need this sort with high complexity and unclear result

belovmd (commented):
Please put tasks descriptions.

dictionary_practice.py
Coromilly: In your example function returns dictionary in alphabetical order. In my function return is the same) if this sort isn't needed i'll correct the program.

dictionary_practice.py
```diff
@@ -0,0 +1,18 @@
...
+    are square of keys. n – function argument. Default is 20.
+    """
+    return {num: num**2 for num in range(1, n + 1)}
+
+
+def count_characters(s):
+    """Function count and return the numbers of each character in a string."""
+    dct = {}
+    for symb in s:
+        if symb in dct:
```
belovmd: we have more generic way without if condition here - please take a look at dict methods

list_practice.py
```diff
@@ -0,0 +1,17 @@
+# 1. Use a list comprehension to construct the list
+# ['ab', 'ac', 'ad', 'bb', 'bc', 'bd'].
+lst1 = [x + y for x in ('a', 'b') for y in ('b', 'c', 'd')]
```
belovmd: can we use 'ab' etc here?

belovmd: can we use 'ab' etc here?

tuple_practice.py
```diff
@@ -0,0 +1,16 @@
+# 1. Create the list ['a', 'b', 'c'], then create a tuple from that list
+lst1 = list('abc')
```
belovmd: do we need this type conversion here?

belovmd (commented):
please address these issues

tuple_practice.py
Coromilly: the way of creating list isn't mentioned.

list_practice.py
Coromilly: solved.

Coromilly: solved.

dictionary_practice.py
Coromilly: solved.

belovmd (commented):
8.5
it's maximal score for full completion of this hw.

PR #61, opened 2019-11-04
Homework3: Aleksandr Strokov

alstrokov:


src/task1.py
```diff
@@ -0,0 +1,23 @@
+"""List practice"""
+
+# 1
+list1 = [ltr1 + ltr2 for ltr1 in ['a', 'b'] for ltr2 in ['b', 'c', 'd']]
```
belovmd: 'ab'?

src/task1.py
```diff
@@ -0,0 +1,23 @@
...
+print(list1)
+
+# 2
+list2 = list1[::2]
+print(list2)
+
+# 3
+list3 = [str(x) + 'a' for x in range(1, 5)]
```
belovmd: do we need range here?

src/task1.py
```diff
@@ -0,0 +1,23 @@
...
+print(list3)
+
+# 4
+print(list3.pop(1))
+print(list3)
+
+# 5
+list4 = list3.copy()
+list4.append('2a')
```
belovmd: it would be in the end

src/task2.py
```diff
@@ -0,0 +1,19 @@
+"""Tuple practice"""
+# 1
+lst = ['a', 'b', 'c']
+tpl = tuple(lst)
+
+# 2
+tpl = ('a', 'b', 'c')
+lst = list(tpl)
+
+print("{}: {}".format(type(lst), lst))
```
belovmd: why do we have this prints but print nothing above? it seems it should be similar

src/task3.py
```diff
@@ -0,0 +1,31 @@
+"""Dictionary practice"""
+
+
+# 1
+def generate_numbers(num=20):
+    dct = dict()
```
belovmd: can we use dict comprehension here?

src/task3.py
```diff
@@ -0,0 +1,31 @@
...
+print(generate_numbers(3))
+print(generate_numbers())
+
+
+# 2
+def count_characters(count_me_string):
+    dct = dict()
+
+    for char in count_me_string:
+        if char in dct.keys():
```
belovmd: we have more native way here. please take a look at dict methods

src/task3.py
```diff
@@ -0,0 +1,31 @@
...
+            dct[char] += 1
+        else:
+            dct[char] = 1
+
+    return dct
+
+
+print(count_characters("abcdefgabc"))
```
belovmd: redundant print

src/task3.py
```diff
@@ -0,0 +1,31 @@
...
+def generate_numbers(num=20):
+    dct = dict()
+
+    for el in range(num + 1):
+        dct[el] = el**2
+
+    return dct
+
+
+print(generate_numbers(3))
```
belovmd: redundant prints

belovmd (commented):
Please add tasks description to files
Please address next issues:

src/task1.py
alstrokov: Yes, 'ab' looks much more accurate

src/task1.py
alstrokov: Is plain string acceptable solution here?

src/task1.py
alstrokov: Yes, in task description it wasn't explicitly declared that `2a` should be added to its original position

alstrokov:
Homework updated and ready for code review

src/task3.py
```diff
@@ -0,0 +1,28 @@
...
+
+# 1
+def generate_numbers(num=20):
+    return {el: el**2 for el in range(num + 1)}
+
+
+# 2
+def count_characters(count_me_string):
+    return {
+        char: count_me_string.count(char)
```
belovmd, line 26: now it's O(n**2)
dct[char] = dct.get(char, 0) +1

belovmd (commented):
7.5

PR #63, opened 2019-11-05
Homework3: Olga Akhmetova

Trylar:
work with data structures
add some easy tasks

### DIFF
--- a/src/homework3/dictionary.py
+++ b/src/homework3/dictionary.py
@@ -1,4 +1,5 @@
 """Dictionary exercises"""
+from collections import Counter
 
 
 # Define a function generate_numbers(number) which returns a dictionary where
@@ -16,4 +17,12 @@ def generate_numbers(n=20):
 # Then, the return result of the function should be:
 # {'a': 2, 'b': 2, 'c': 2, 'd': 1, 'e': 1, 'f': 1, 'g': 1}
 def count_characters(count_me_string):
-    return dict([(ltr, count_me_string.count(ltr)) for ltr in count_me_string])
+    dct = {}
+    for ltr in count_me_string:
+        dct[ltr] = dct.get(ltr, 0) + 1
+    return dct
+
+
+# second version
+def count_characters2(count_me_string):
+    return dict(Counter(count_me_string))
--- a/src/homework3/dictionary.py
+++ b/src/homework3/dictionary.py
@@ -1,0 +1,19 @@
+"""Dictionary exercises"""
+
+
+# Define a function generate_numbers(number) which returns a dictionary where
+# the keys are numbers between 1 and n (both included) and
+# the values are square of keys. n – function argument. Default is 20.
+def generate_numbers(n=20):
+    return dict([(i, i * i) for i in range(1, n + 1)])
+
+
+# Define a function count_characters(count_me_string) which count and return
+# the numbers of each character in a count_me_string argument.
+# Example:
+# If the following string is given as argument to the function:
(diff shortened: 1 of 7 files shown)


src/homework3/list.py
```diff
@@ -0,0 +1,24 @@
+"""Play around  with lists"""
+
+# Use a list comprehension to construct the list
+# ['ab', 'ac', 'ad', 'bb', 'bc', 'bd'].
+list1 = [ltr + ltr2 for ltr in "ab" for ltr2 in "bcd"]
```
belovmd, line 5: 👍

src/homework3/list.py
```diff
@@ -0,0 +1,24 @@
...
+print(list1)
+
+# Use a slice on the above list to construct the list ['ab', 'ad', 'bc'].
+list2 = list1[::2]
+print(list2)
+
+# Use a list comprehension to construct the list ['1a', '2a', '3a', '4a'].
+list3 = [str(dgt) + "a" for dgt in range(1, 5)]
```
belovmd: we have more native way instead of range

src/homework3/dictionary.py
```diff
@@ -0,0 +1,19 @@
...
+
+# Define a function count_characters(count_me_string) which count and return
+# the numbers of each character in a count_me_string argument.
+# Example:
+# If the following string is given as argument to the function:
+# abcdefgabc
+# Then, the return result of the function should be:
+# {'a': 2, 'b': 2, 'c': 2, 'd': 1, 'e': 1, 'f': 1, 'g': 1}
+def count_characters(count_me_string):
+    return dict([(ltr, count_me_string.count(ltr)) for ltr in count_me_string])
```
belovmd: it has O(n**2) complexity
we has O(n) solution if we'll use some dict methods

belovmd (commented):
Thank you for your work!
Please address issues according hometask tasks
I will review rest of your tasks later - tomorrow or on Wednesday to give feedback.

belovmd (commented):
10

src/homework3/dictionary.py
```diff
@@ -0,0 +1,28 @@
+"""Dictionary exercises"""
+from collections import Counter
+
+
+# Define a function generate_numbers(number) which returns a dictionary where
+# the keys are numbers between 1 and n (both included) and
+# the values are square of keys. n – function argument. Default is 20.
+def generate_numbers(n=20):
+    return dict([(i, i * i) for i in range(1, n + 1)])
```
lavoncyk, line 9: `[]` are not necessary inside `dict()` call. It would be enough to pass generator expression there:
```py
dict((i, i * i) for i in range(1, n + 1))
```

src/homework3/timus/a_whole_lot_of_walnuts.py
```diff
@@ -0,0 +1,48 @@
...
+    number, result = input().split()
+    number = int(number)
+    if experiments[number] and experiments[number] != result:
+        print("Inconsistent")
+        break
+    else:
+        experiments[number] = result
+else:
+    max_hungry = max(
+        [key for key in experiments if experiments[key] == HUNGRY])
```
lavoncyk, line 42: `[]` are not necessary inside `max()` or `min()` calls. It would be enough to pass generator expression there:
```py
max(key for key in experiments if experiments[key] == HUNGRY)
```

lavoncyk (approved):

belovmd:
Thank you Alexandr

PR #70, opened 2019-11-08
Homework3: Andrey Kapitonov

akapitonov1999:


Home3.py
```diff
@@ -0,0 +1,24 @@
+"""List practice"""
+lst = [ltr + ltr2 for ltr in "ab" for ltr2 in "bcd"]
+lst = lst[::2]
+new_list = ['1a', '2a', '3a', '4a']
```
belovmd: it should be list comprehension as I remember

Home3.py
```diff
@@ -0,0 +1,24 @@
...
+print(len(tpl4))
+
+
+# Dictionary practice
+def generate_numbers(number=20):
+    return {dct: dct ** 2 for dct in range(1, number + 1)}
+
+
+def count_characters(count_me_string):
+    return {dict_: count_me_string.count(dict_) for dict_ in count_me_string}
```
belovmd: 1. good naming is symbol or element not dict_ because it's not a dict
2. It has O(n**2) complexity. There is O(n) solution if we'll use dict methods

belovmd (commented):
1. Please add tasks description
2. Please address these issues:

Home3.py
```diff
@@ -0,0 +1,28 @@
...
+# Dictionary practice
+def generate_numbers(number=20):
+    return {dct: dct ** 2 for dct in range(1, number + 1)}
+
+
+def count_characters(count_me_string):
+    dct = {}
+    dct = dct.fromkeys(count_me_string, 0)
+    for i in count_me_string:
+        dct[i] = dct[i]+1
```
belovmd, line 27: .get(i, 0) would be more pythonic

belovmd (commented):
7.5 because of failed code style

PR #71, opened 2019-11-09
Homework3: Sergey Streltsov

Mantikor:


src/homework03.py
```diff
@@ -0,0 +1,62 @@
...
+
+    """
+    return {n: n * n for n in range(1, number + 1)}
+
+
+def count_characters(count_me_string):
+    """Dictionary practice
+
+    """
+    return {ch: count_me_string.count(ch) for ch in count_me_string}
```
belovmd: it has O(n**2) complexity. 
There is O(n) solution - please take a look at dicts methods

src/homework03.py
```diff
@@ -0,0 +1,62 @@
...
+
+def list_practice():
+    """List practice
+
+    """
+    src_lst = [i + j for i in 'ab' for j in 'bcd']
+    print(src_lst)
+    new_lst = src_lst[::2]
+    print(new_lst)
+    sec_lst = [str(i) + 'a' for i in range(1, 5)]
```
belovmd: there is more native way instead of range

belovmd (commented):
1. Please add tasks description
2. Please address these minor issues:

Mantikor:
Fixes for all remarks

belovmd (commented):
8.5
it's maximal score for full completion of this hw.

PR #72, opened 2019-11-10
Homework3: Pavel Ralko

par216:


H3-1.py
```diff
@@ -0,0 +1,27 @@
+'''
```
belovmd: usually we use """

H3-2.py
```diff
@@ -0,0 +1,25 @@
+'''
```
belovmd: """

H3-3.py
```diff
@@ -0,0 +1,22 @@
...
+Dictionary practice
+'''
+
+
+# Define a function generate_numbers(number)
+# which returns a dictionary where the keys are numbers
+# between 1 and n (both included) and the values are square
+# of keys. n–function argument. Default is 20.
+def generate_numbers(number):
+    return dict((i, i ** 2) for i in range(number))
```
belovmd: please use dict comprehension

H3-3.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+# Define a function count_characters(count_me_string)
+# which count and return the numbers
+# of each character in a count_me_string argument.
+def count_characters(count_me_string):
+    return {i: count_me_string.count(i) for i in count_me_string}
```
belovmd: it would have O(n**2) complexity. 
we have O(n) solution

belovmd (changes requested):
Please address these issues:

H3-1.py
```diff
@@ -0,0 +1,27 @@
...
+print(list1)
+
+# Use a slice on the above list to construct
+# the list ['ab', 'ad', 'bc'].
+list2 = list1[::2]
+print(list2)
+
+# Use a list comprehension to construct
+# the list ['1a', '2a', '3a', '4a'].
+list3 = [e1 + e2 for e1 in '1234' for e2 in 'a']
```
belovmd, line 16: we can just print e1+'a'

belovmd (commented):
8

8.5 is maximal score for full completion of this hw.

PR #74, opened 2019-11-12
Homework4: Anton Stychnevsky

Stychnevsky:
Homework # 4

### DIFF
--- a/HW4 - task1.py
+++ b/HW4 - task1.py
@@ -8,7 +8,33 @@ from time import sleep
 from time import strftime
 
 
-def func_calls_history(func):
+def func_calls_history_v1(func):
+    """decorator which save history for all functions
+
+    There is decision for first version of task - decorator save all functions
+    called during program running
+    """
+    if not hasattr(func_calls_history_v1, 'calls_history'):
+        func_calls_history_v1.calls_history = []
+
+    def wrapper(*args, **kwargs):
+        func_call_result = func(*args, **kwargs)
+        func_calls_history_v1.calls_history.append({
+            'func': func.__name__,
+            'end_time': strftime("%Y-%m-%d %T", gmtime()),
+            'result': func_call_result
+            })
+        sleep(1)
+        return func_call_result
+    return wrapper
+
+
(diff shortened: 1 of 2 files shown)


HW4 - task1.py
```diff
@@ -0,0 +1,22 @@
...
+
+def func_calls_history(func):
+    calls_history = []  # don't use nonlocal in wrapper because list is mutable
+
+    def wrapper(*args, **kwargs):
+        func_call_result = func(*args, **kwargs)
+        calls_history.append({'func': func.__name__,
+                              'end_time': strftime("%Y-%m-%d %T", gmtime()),
+                              'result': func_call_result
+                              })
```
Stychnevsky, line 45: There is frequent situation when we have some function\input which return values(more than one and similar in every call), and this function\input is called unknown number of times (like "calls_history" in my program). What is the best type to save such sequences? I used list of dicts, but may be list of lists or list of tuples is better? Or may be it's better to create a class?

HW4 - task2.py
```diff
@@ -0,0 +1,21 @@
...
+ get_ranges([4,7,10]) // "4,7,10"
+ get_ranges([2, 3, 8, 9]) // "2-3,8-9"
+"""
+
+
+def get_ranges(lst):
+    if len(lst) == 1:
+        return str(lst[0])
+
+    new_lst = ''
```
belovmd: why variable with name list keeps string?

HW4 - task1.py
```diff
@@ -0,0 +1,48 @@
...
+from time import strftime
+
+
+def func_calls_history_v1(func):
+    """decorator which save history for all functions
+
+    There is decision for first version of task - decorator save all functions
+    called during program running
+    """
+    if not hasattr(func_calls_history_v1, 'calls_history'):
```
belovmd, line 17: good solution

belovmd (commented):

HW4 - task1.py
belovmd: you solution is OK

belovmd (commented):
9

PR #75, opened 2019-11-12
Homework4: Aleksandr Strokov

alstrokov:


src/task1.py
```diff
@@ -0,0 +1,40 @@
+"""task 1
+
+Создайте декоратор, который хранит в себе результаты вызовов функций
+и время получения результата за время запуска программы. """
+
+import time
+
+# global dictionary to sotre time and result of decorated functions runs.
+decorator_run_statistic = dict()
```
belovmd, line 9: why don't we use a = {}

src/task2.py
```diff
@@ -0,0 +1,36 @@
...
+            curr = x
+        else:
+            append_range(rez, start, curr)
+            start = x
+            curr = x
+    append_range(rez, start, curr)
+    return rez
+
+
+print(get_ranges([0, 1, 2, 3, 4, 7, 8, 10]))
```
belovmd: wrong output format - it marked after // in task description
overall it's good solution

belovmd (commented):

alstrokov:
Ready to be reviewed

src/task1.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+# decorator
+def run_and_save_decorator(funct):
+    global decorator_run_statistic
+
+    def wrapper(*args, **kwargs):
+        rez = funct(*args, **kwargs)
+        time_run = time.time()
+        decorator_run_statistic[time_run] = rez
```
belovmd, line 19: how can I detect which function was called?

src/task2.py
```diff
@@ -0,0 +1,46 @@
...
+
+    rez = list()
+    start = lst[0]
+    curr = lst[0]
+
+    for x in lst:
+        if x - curr <= 1:
+            curr = x
+        else:
+            append_range(rez, start, curr)
```
belovmd, line 37: it's not best idea to change mutable inside function. we can just list.append(func) for example

belovmd (approved):
8

PR #76, opened 2019-11-12
Homework3: Nikolai Potovin

adnike:


belovmd (commented):
it's not 3rd hw solution

adnike:
Sorry, sent 4th hw somehow. Should be 3rd now

src/hw3.py
```diff
@@ -0,0 +1,72 @@
...
+# 2 Define a function count_characters(count_me_string)
+# which count and return the numbers of each character
+# in a count_me_string argument.
+# If we need to count 'a' and 'A' separately then remove the 'string' variable
+# and use count_me in cycle both times instead
+def count_characters(count_me):
+    """count the number of characters in the string (case non-sensitive)"""
+    dict3_2 = {}
+    for let in count_me.lower():
+        if let in dict3_2:
```
belovmd, line 65: we can use get here to avoid this in condition

belovmd (commented):
8

PR #77, opened 2019-11-12
Homework4: Olga Akhmetova

Trylar:


src/homework4/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+
+    return wrapper
+
+
+@store
+def pow(a, b):
+    return a ** b
+
+
+for i, j in zip(range(10 ** 2), range(10 ** 3)):
```
belovmd: can we use list comprehension [(i, j) for i in range for j in range] here? 
pow(*list_element)
zip is not usually used in python as all this functionality can be handled by comprehension or decorators.

src/homework4/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def wrapper(*args, **kwargs):
+        if args in result2:
+            return result2[args].result
+        else:
+            start = time.time()
+            for i in range(1000):  # to make time bigger
+                res = f(*args, **kwargs)
+            stop = time.time()
+            result2[args] = {"result": res % (10 ** 8), "time": (stop - start)}
```
belovmd: why do we need this division?

src/homework4/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+# another solution, sort of cache
+def store2(f):
+    result2 = {}
+
+    def wrapper(*args, **kwargs):
+        if args in result2:
+            return result2[args].result
+        else:
+            start = time.time()
+            for i in range(1000):  # to make time bigger
```
belovmd: it doesn't look as a good idea because if function has some random elements it would last for different time.

src/homework4/functions.py
```diff
@@ -0,0 +1,45 @@
...
+    return res_lst
+
+
+# second solution, works with all numbers
+def get_ranges2(int_list):
+    res = []
+    length = len(int_list)
+    l = r = 0
+    while l < length:
+        while r < length - 1 and int_list[r] + 1 == int_list[r + 1]:
```
belovmd: we can move this to some find function to make our code more lightweight.

src/homework4/functions.py
```diff
@@ -0,0 +1,45 @@
...
+    else:
+        res_lst.append(str(num))
+    return res_lst
+
+
+# second solution, works with all numbers
+def get_ranges2(int_list):
+    res = []
+    length = len(int_list)
+    l = r = 0
```
belovmd: please name these variables accordingly

src/homework4/functions.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+# first solution, works only with positive numbers
+def get_ranges(int_list):
+    return ",".join(reduce(gluing, int_list, []))
+
+
+def gluing(res_lst, num):
+    if res_lst:
+        last = res_lst[-1].rpartition("-")[-1]
```
belovmd: I've studied rpartition function 🧪

belovmd (commented):
Great. 

src/homework4/decorators.py
Trylar: division just for more readable output

src/homework4/functions.py
Trylar: here it works like split, I unnecessarily complicated

src/homework4/decorators.py
```diff
@@ -0,0 +1,49 @@
...
+
+result = []
+
+
+def store(f):
+    def wrapper(*args, **kwargs):
+        start = time.time()
+        res = f(*args, **kwargs)
+        stop = time.time()
+        result.append({"result": res, "time": (stop - start)})
```
belovmd, line 13: here we can't detect which function was called

belovmd (approved):
9

PR #78, opened 2019-11-12
Homework4: Anton Koliada

JustgoodDeal:


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "python.linting.pylintEnabled": false,
```
belovmd: if you use vscode please update gitignore

Homework4/Task1.py
```diff
@@ -0,0 +1,40 @@
...
+        global results
+        results = {}
+
+    def wrapper(*args, **kwargs):
+        before_call = datetime.datetime.now()
+        res = func(*args, **kwargs)
+        after_call = datetime.datetime.now()
+        time = after_call - before_call
+        name = func.__name__
+        results[name] = results.get(name, []) + [(res, str(time))]
```
belovmd: it looks append would be more native here

Homework4/Task1.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    if 'results' not in globals():
+        global results
+        results = {}
+
+    def wrapper(*args, **kwargs):
+        before_call = datetime.datetime.now()
+        res = func(*args, **kwargs)
+        after_call = datetime.datetime.now()
+        time = after_call - before_call
```
belovmd: as I remember it would be datetime.timedelta. it would need conversion and would return formatted time. overall it's ok. alternatively you can use from time import time

Homework4/Task2.py
```diff
@@ -0,0 +1,20 @@
+def get_ranges(numbers):
+    """Curtails non-empty list of non-repeating integers
+
+    sorted in ascending order
+    """
+    curtail_numbers = ''
+    index = 0
+    while index < len(numbers) - 1:
```
belovmd: good solution! 
the only thing it looks a bit heavyweight and maybe we can refactor it a bit 
(i.e. rename vars, prepare strings to concat in another line... (it's not usually used a += expression but a += value))

belovmd (commented):
Thank you for your work. please address these minor issues.

belovmd (commented):
9

PR #79, opened 2019-11-13
Homework4: Sergey Streltsov

Mantikor:


src/homework04.py
```diff
@@ -0,0 +1,77 @@
...
+import time
+
+
+def save_history(f_name):
+    """History decorator
+
+    Create a decorator that stores the results of function calls and the
+    time it takes to get the result during program startup
+    """
+    history = list()
```
belovmd, line 24: alternatively history = []

src/homework04.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def wrapper(*args, **kwargs):
+        start_time = int(round(time.time() * 1000))
```
belovmd: why can't we just use time.time and if we need some conversion than make it in the end?

src/homework04.py
```diff
@@ -0,0 +1,77 @@
...
+        f_result = f_name(*args, **kwargs)
+        stop_time = int(round(time.time() * 1000))
+        est_time = stop_time - start_time
+        history.append({'name': f_name.__name__,
+                        'result': f_result,
+                        'time_ms': est_time})
+        print(history)
```
belovmd: why should decorator print anything?

src/homework04.py
```diff
@@ -0,0 +1,77 @@
...
+                last = number
+            else:
+                raw_str += ',' + str(number)
+                last = number
+        else:
+            raw_str += str(number)
+            last = number
+    new_list = list()
+    for block in raw_str.split(','):
+        if block.count('-') > 1:
```
belovmd, line 69: it looks like O(n**2) - bad complexity

src/homework04.py
```diff
@@ -0,0 +1,77 @@
...
+            new_block = '-'.join([block.split('-')[0], block.split('-')[-1]])
```
belovmd, line 70: Overall it looks like here we make redundant structure like num-num+1-num+2, other_num and than parse it with sometimes O(n**2) complexity. 
It looks it's exceptionally redundant as we have a lot of implicit parts (why do we need all these -, why do we need 0 and -1 elements. ) 
Overall, it looks too complex for this task.

belovmd (commented):
please take a look at these issues:

src/homework04.py
Mantikor: added

src/homework04.py
Mantikor: ok, I agree, moved conversion to the end

src/homework04.py
Mantikor: for viewing results of work ) fixed

src/homework04.py
Mantikor: fixed

src/homework04.py
Mantikor: the structure was: num1-num2-num3,num4,num5,num6-num7.... ect. And after I parse the string and eliminate body from blocks between commas. Now I implemented v2 of algo, but I mean this v2 not so pythonic ) because using list indexes...

src/homework04.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get_ranges_v2(in_list):
+    """Collapsed list v2
+
+    Implement the get_ranges function which receives a non-empty list
+    of non-repeating integers, sorted in ascending order, which this
+    list “collapses”
+    """
+    out_str = ''
+    for n in range(len(in_list) - 1):
```
belovmd, line 86: enumerate

belovmd (approved):
9

PR #80, opened 2019-11-13
Homework4: Pavel Martysiuk

PavelMartysiuk:


src/decor.py
```diff
@@ -0,0 +1,34 @@
...
+from time import time
+
+data = defaultdict(list)
+
+
+def storage(func):
+    def wrapper(*args, **kwargs):
+        start_time = time()
+        res = func(*args, **kwargs)
+        data[func.__name__].extend([res, start_time - time()])
```
belovmd, line 14: would it work with from decor import storage?

src/get_ranges.py
```diff
@@ -0,0 +1,25 @@
+"""Реализовать функцию get_ranges, которая получает на вхох непустой список
```
belovmd: вход misprint

src/get_ranges.py
```diff
@@ -0,0 +1,25 @@
...
+неповторяющихся целых чисел, отсортированных по возрастанию, которая этот
+список сворачивает"""
+
+
+def get_ranges(lst):
+    out_str = ''
+    for el in range(len(lst)):
```
belovmd: maybe it's better to add additional check instead of try-except block?

belovmd (commented):

src/get_ranges.py
PavelMartysiuk: One of the principales of python is  ask forgiveness not permission. Because I think that using the try-except block is better.

belovmd: It can be in case of 2 conditions:
1. You know where and how would it be handled.
2. It's not attempt to hide error in algorithm.
In your case you wrote algorithm that would normally throw this exception. Exceptions are for handling exceptions, not a part of algorithm.

belovmd (approved):
9

PR #81, opened 2019-11-13
Homework4: Andrey Kuksik

Kuxsoger:


src/homework4/decorator.py
```diff
@@ -0,0 +1,18 @@
+"""
+Создайте декоратор, который хранит в себе результаты вызовов функций
+и время получения результата за время запуска программы.
+"""
+
+import time
+
+
+def log_results(lst):
```
belovmd: It's a good idea but I want just to decorate function, not create other variables etc.

src/homework4/get_ranges.py
```diff
@@ -0,0 +1,22 @@
...
+которая этот список “сворачивает”
+"""
+
+
+def get_ranges(lst):
+    result = ""
+    start, prev = lst[0], lst[0]
+    for num in lst[1:]:
+        if num - prev != 1:
+            result += str(start) + ('-' + str(prev)) * (prev != start) + ','
```
belovmd: here it's not the best solution as it looks too heavyweight

belovmd (commented):
Thank you for your work please address these issues:

src/homework4/decorator.py
```diff
@@ -0,0 +1,20 @@
...
+Создайте декоратор, который хранит в себе результаты вызовов функций
+и время получения результата за время запуска программы.
+"""
+
+import time
+
+
+def log_results(func):
+    if not hasattr(log_results, "logs"):
+        log_results.logs = []
```
belovmd, line 11: why can't we use logs = []?

belovmd (approved):
8

src/homework4/decorator.py
belovmd: i will improve score if you can explain why do we need hasattr here. you can email explanation.

Kuxsoger: i sent a message to python.epam.q42019@gmail.com

belovmd:
thank you. 9;

PR #82, opened 2019-11-13
Homework4: Aleksandr Levdansky

Coromilly:


func_decor.py
```diff
@@ -0,0 +1,64 @@
+from time import time
+
+registry = {}
+
+
+def wrapper(func):
```
belovmd: usually we call wrapper internal function

func_decor.py
```diff
@@ -0,0 +1,64 @@
...
+def get_ranges(lst):
+    """Function "get_ranges".
+
+    This function takes list of unique sorted ascending integer numbers
+    and returns this list rolled up.
+    For example:
+    get_ranges([0, 1, 2, 3, 4, 7, 8, 10]) // "0-4,7-8,10"
+    """
+    pos = 0
+    rolled_list = str(lst[0])
```
belovmd: am I correct you name string variable list?

func_decor.py
```diff
@@ -0,0 +1,64 @@
...
+    return timer
+
+
+@wrapper
+def spam(*args):
+    return {arg: arg ** 2 for arg in args}  # spam = wrapper(spam)
+
+
+@wrapper
+def get_ranges(lst):
```
belovmd, line 25: your function doesn't work correctly and has wrong output

belovmd (commented):
second task doesn't work correctly. 
Also please address other minor issues

func_decor.py
Coromilly: Yep. Y answered in telegram some days ago that output should be a string. The task was to roll out the list.
I'll correct to "roll".

func_decor.py
Coromilly: solved.

func_decor.py
Coromilly: solved.

func_decor.py
```diff
@@ -0,0 +1,67 @@
...
+
+registry = {}
+
+
+def timer(func):
+    def wrapper(*args, **kwargs):
+        start = time()
+        res = func(*args, **kwargs)
+        stop = time()
+        registry[func.__name__] = (
```
belovmd, line 11: it would rewrite results after every function call

get_ranges.py
```diff
@@ -0,0 +1,26 @@
+def get_ranges(lst):
```
belovmd, line 1: why do we need two same functions? can't we use import?

belovmd (commented):
6.5

PR #86, opened 2019-11-15
Homework5: Olga Akhmetova

Trylar:


src/homework5/1/runner.py
```diff
@@ -0,0 +1,12 @@
+import inspect
+import pytasks
+
+
+def runner(*args):
+    if not args:
+        func_tuples = inspect.getmembers(pytasks, inspect.isfunction)
```
belovmd: 🧪 
```suggestion
functions = [element for element in dir(pytasks) if not element.startswith(__) and callable(element)]
```

belovmd (commented):
👍 good job! please take a look at this suggestion. I'll review your pr later

src/homework5/1/runner.py
Trylar: I changed code to more clear; dir returns string that is not callable, so check is performed later.

src/homework5/2/imdb.py
```diff
@@ -0,0 +1,56 @@
...
+Find list of Top250 movies. Extract  titles information.
+Create 3 text files top250_movies.txt – contains titles, ratings.txt –
+contains histogram for rating, years.txt – contains histogram for years.
+Don’t add ratings.list file to PR, but add your created 3 files"""
+from collections import Counter
+import os
+
+
+def count_and_write_to_file(filename, column):
+    data = Counter([movie[column] for movie in movies_top250])
```
belovmd: it's not very good to work with global var (top250) expecially if it's defined after this function. you can pass it as parameter

belovmd (commented):

belovmd (commented):
10

PR #89, opened 2019-11-16
Homework5: Anton Stychnevsky 

Stychnevsky:
Homework # 5


imdb.py
```diff
@@ -0,0 +1,99 @@
...
+
+head = [next(ratings_file) for x in range(28)]
+titles = []
+years = []
+rates = []
+for rows in range(250):
+    row = next(ratings_file)
+    rates.append(row.split()[2])
+    titles.append(' '.join(row.split()[3:]))
+    years.append(row.split()[-1].strip('()'))
```
Stychnevsky: Is it pythonic to use 4-5 methods in a row (in one line)? Or better separate it to two lines?

imdb.py
```diff
@@ -0,0 +1,99 @@
...
+        elements_dict[el] = elements_dict.get(el, 0) + 1
+    return elements_dict
+
+
+def nice_hat_for_years_file(years_file, counter=True):
+    years_file.write('Year | Amount  |' + counter * 'A' + '\n')
+    years_file.write('_____|_________|' + counter * '_' + '\n')
+
+
+def create_histogram(lst, filename='histogram.txt', vertical=True,
```
belovmd: this functions looks to big it would be better to split it into two (create vertical and usual) and refactor a bit

imdb.py
```diff
@@ -0,0 +1,99 @@
...
+                          '#' * dct[key] +
+                          ' ' * (max_value - dct[key]) +
+                          '|' + str(dct[key]) * counter)
+        for row1 in matrix:
+            histogram_file.write(row1 + '\n')
+    else:
+        for key in sorted(dct.keys()):
+            matrix.append(' ' * (max_value - dct[key]) + '#' * dct[key])
+            keys_list.append(str(key) + ' ' * (max_key_len - len(str(key))))
+            ziped = zip(*matrix)
```
belovmd: i don't understand the purpose of this zip
it seems to me we can avoid it

imdb.py
belovmd: we usually use [line.strip for line in lines.split] it's ok. 
but this append looks weird for me and it seems it should be refactored. 
also please avoid magic numbers (2 3)

runner.py
```diff
@@ -0,0 +1,19 @@
...
+    results to screen.
+    runner() calls all functions from pytasks
+    """
+
+    if not functions:
+        functions = dir(pytasks)
+    for func in functions:
+        if not func.startswith('__'):
+            if hasattr(pytasks, func):
+                print('{}: {}'.format(func, getattr(pytasks, func)()))
```
belovmd, line 17: it seems we should check if this func callable

runner.py
```diff
@@ -0,0 +1,19 @@
...
+def runner(*functions):
+    """runner
+
+    Called function(s) from pytasks module with default values and printed
+    results to screen.
+    runner() calls all functions from pytasks
+    """
+
+    if not functions:
+        functions = dir(pytasks)
```
belovmd: we can filter (starts with) it here

belovmd (commented):

imdb.py
```diff
@@ -0,0 +1,96 @@
...
+    titles.append(' '.join(split_row[titles_column_num: years_column_num]))
+    years.append(split_row[years_column_num].strip('()'))
+ratings_file.close()
+
+os.chdir(project_directory)
+titles_file = open("top250_movies.txt", "w")
+for film_title in titles:
+    titles_file.write(film_title + '\n')
+titles_file.close()
+years = map(lambda year: year[:4], years)
```
belovmd, line 94: [year[:4] for year in years}]

belovmd (commented):
9

PR #92, opened 2019-11-16
 Homework4: Anastasiya Infarovich

AnInf:
Homework 4

### DIFF
--- a/src/task4/1.py
+++ b/src/task4/1.py
@@ -1,0 +1,34 @@
+"""Decorator that keeps result of function and its execution time"""
+
+import time
+
+
+def timer(f):
+    def wrapper(*args, **kwargs):
+        t0 = time.time()
+        res = f(*args, **kwargs)
+        t1 = time.time()
+        function_execution_time = t1 - t0
+        print(function_execution_time)
+        return res
+    return wrapper
+
+
+@timer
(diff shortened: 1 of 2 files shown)


src/task4/1.py
```diff
@@ -0,0 +1,34 @@
...
+import time
+
+
+def timer(f):
+    def wrapper(*args, **kwargs):
+        t0 = time.time()
+        res = f(*args, **kwargs)
+        t1 = time.time()
+        function_execution_time = t1 - t0
+        print(function_execution_time)
```
belovmd, line 12: it should keep time and result not print

src/task4/2.py
```diff
@@ -0,0 +1,29 @@
+"""Function, that receives not empty list of non-repeating integers,
+ sorted descending and "shortens" the list"""
+
+
+def get_ranges(my_list):
+    a = b = None
```
belovmd, line 6: overall it's good solution but naming like `a b i` make it completely not understandable

belovmd, line 6: can we avoid this none if we'll assign them in our first iteration?

belovmd (commented):
please address these issues

belovmd (commented):
please take a look at previous review
5,5

PR #103, opened 2019-11-18
Homework6: Anton Stychnevsky 

Stychnevsky:
Homework # 6

### DIFF
--- a/HW6.py
+++ b/HW6.py
@@ -1,3 +1,5 @@
+from abc import ABC
+from abc import abstractmethod
 import random
 """Homework 6 - Kitchen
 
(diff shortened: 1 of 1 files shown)


HW6.py
```diff
@@ -0,0 +1,224 @@
...
+            self.skill += 1
+        else:
+            print('{self.master} don\'t have enough money to study')
+
+
+class Product(object):
+    """Product is some food we can eat or put to refrigerator
+
+    It can be cold and freeze type.
+    Every product aso have size -oit shows how many cells this product need
```
belovmd: it 
misprint

HW6.py
```diff
@@ -0,0 +1,224 @@
...
+    money = 100
+
+    def __init__(self, skill, name='Master', company='own'):
+        self.skill = skill
+        self.name = name
+        self.company = company
+
+    def advertisement(self):
+        """Advertising of master. Skills always increase by two"""
+        print('Highly skilled master {0} from {1} company is looking for work!'
```
belovmd: i would prefer key value args here

belovmd: i would prefer key value args here

HW6.py
```diff
@@ -0,0 +1,224 @@
...
+              ' and {3} free freeze cells '.format(type(self).__name__,
+                                                   self.name,
+                                                   self.cold_cells,
+                                                   self.freeze_cells))
+
+
+class RefrigeratorAtlant(Refrigerator):
+    """Belorussian refrigerator with small capacity"""
+    made_in = 'Belarus'
+    freeze_cam_capacity = 4
```
belovmd: i don't understand why it doesn't depend on model of fridge

HW6.py
```diff
@@ -0,0 +1,224 @@
...
+        self.pr_type = pr_type
+        self.size = size
+
+
+class Refrigerator(object):
+    """Refrigerator is place where we can save products
+
+    New refrigerator always is empty and able to work
+    """
+    freeze_cells = 0
```
belovmd: Am I correct if we'll have to fridges they would have same cells num

HW6.py
```diff
@@ -0,0 +1,224 @@
...
+        print('He fixed it' if job_result else 'He can\'t fix it')
+
+    def earn_money(self, salary=200):
+        """Parent get some money after this method"""
+        self.money += salary
+
+
+class Child(User):
+    """Child is user who can hit refrigerator"""
+    def hit_fridge(self, refr):
```
belovmd, line 106: maybe it should be added to get or put food method for child?

HW6.py
```diff
@@ -0,0 +1,224 @@
...
+            print('{self.master} don\'t have enough money to study')
+
+
+class Product(object):
+    """Product is some food we can eat or put to refrigerator
+
+    It can be cold and freeze type.
+    Every product aso have size -oit shows how many cells this product need
+    """
+    in_cold_cam = None
```
belovmd: When I'll put products would be created instance variable for each instance. 
the only use of these vars are not to take exception in your if-elif condition. it should be refactored.

HW6.py
```diff
@@ -0,0 +1,224 @@
...
+        if refr.made_in == 'The Netherlands':
+            if self.skill >= 7:
+                refr.broken = False
+                customer.money -= self.skill * 100 * 1.5
+                return True
+        elif refr.made_in == 'Belarus':
+            if self.skill >= 3:
+                refr.broken = False
+                customer.money -= self.skill * 100 * 1
+                self.skill += 0.5
```
belovmd: he can unlimited improve his skill repairing atlants and can't level up when repair netherlands.

belovmd (commented):
overall it's too many elif as for me. 
also please take a look at these issues:

HW6.py
Stychnevsky: But child have "get" and "put" methods, because class Child is inheritor of class User, and class User have those methods

HW6.py
Stychnevsky: Yes, because zero mean that it is zero occupied cells in new refrigerator.

I rename this variable to "occupied_cells" to make it clearer

HW6.py
Stychnevsky: Fixed.
Now every company has own models row. And we can change model or create new using "create_or_update_model" class method

HW6.py
Stychnevsky: Fixed.
Now instead of "in_cold_cam" and "in_freeze_cam" it is only one product property - location.

Stychnevsky:
> overall it's too many elif as for me.

Fixed. Now it is no elifs in program. I did it using:
1) Dicts instead of two property variables, for example: capacity = {'freeze': 4,'cold': 6} instead of freeze_cam_capacity = 4 and cold_camp_capacity = 6
2) Abstract class Refrigerator
3) In first version of program "cold" products had to be put in freeze camera if it is no place in cold camera. This fact highly complicated the logic of the model and is not very close to life. So, I decided to make it easier. Now "cold" products can be put only in cold camera,  and "freeze" product only in freeze camera

HW6.py
Stychnevsky: When I invented I mean that Belorussian refrigerators are so bad, that every time they have new problem and master can improve his skills solving it :)
But I agree that is is not logical. Now fixed: Netherlands refrigerators also help to level up. And skills limit has been set.

HW6.py
belovmd: we can reload them, call super and add private _hit method to them I meant

HW6.py
belovmd: I have new Atlant fridge and really like it ^)

belovmd (commented):
8,5

PR #104, opened 2019-11-18
Homework6: Sergey Streltsov

Mantikor:
The simplest room object, but I think this is enough

### DIFF
--- a/src/homework06.py
+++ b/src/homework06.py
@@ -8,10 +8,14 @@ Sergey Streltsov 2019-11-17
 Simplest object of a Room
 """
 
+from datetime import datetime
 import hashlib
 
 
 class Room(object):
+    """Simple room object
+
+    """
 
     def __init__(self, room_number, bed_number, is_lux=False):
         self._room_number = room_number
@@ -20,6 +24,13 @@ class Room(object):
         self._booking_history = dict()
 
     def book_room(self, checkin_date, checkout_date):
+        """Room booking
+
+        :param checkin_date: checkin date
+        :param checkout_date: checkout date
+        :return: confirmation number (hash) or -1 if not
+        available
+        """
         if self.check_booking_interval(
                 checkin_date=checkin_date,
                 checkout_date=checkout_date) and not self.is_booked():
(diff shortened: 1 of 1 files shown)


src/homework06.py
```diff
@@ -0,0 +1,73 @@
...
+        if confirmation_number in self._booking_history.keys():
+            dates = self._booking_history.pop(confirmation_number)
+            print('{}, booking canceled'.format(dates))
+        else:
+            print('Error: booking with conf. number {} not found'.format(
+                confirmation_number))
+
+    def check_booking_interval(self, checkin_date, checkout_date):
+        # check is interval valid
+        return bool(checkin_date and checkout_date)
```
belovmd: you check only two dates here. more of that they still can be invalid if checkout less than check in. also  it can be booked between them. this method completely unclear to me.

src/homework06.py
```diff
@@ -0,0 +1,73 @@
...
+    def cancel_booking(self, confirmation_number):
+        if confirmation_number in self._booking_history.keys():
+            dates = self._booking_history.pop(confirmation_number)
+            print('{}, booking canceled'.format(dates))
+        else:
+            print('Error: booking with conf. number {} not found'.format(
+                confirmation_number))
+
+    def check_booking_interval(self, checkin_date, checkout_date):
+        # check is interval valid
```
belovmd: we use """ for docstrings

src/homework06.py
```diff
@@ -0,0 +1,73 @@
...
+        return bool(checkin_date and checkout_date)
+
+    def is_booked(self, mock_obj=False):
+        # check is booked
```
belovmd: """

src/homework06.py
```diff
@@ -0,0 +1,73 @@
...
+        return mock_obj
```
belovmd, line 79: it's unclear for me. more of that we usually use mocks for tests

src/homework06.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, room_number, bed_number, is_lux=False):
+        self._room_number = room_number
+        self._bed_number = bed_number
+        self._is_lux = is_lux
+        self._booking_history = dict()
+
+    def book_room(self, checkin_date, checkout_date):
+        if self.check_booking_interval(
+                checkin_date=checkin_date,
+                checkout_date=checkout_date) and not self.is_booked():
```
belovmd, line 36: your is booked would be alvais false

belovmd (commented):
please put docstrings to class and methods. 
please address these issues:
it seems for me booking not works as it should

src/homework06.py
Mantikor: yes, but i don't have time for implement logic for is_booking check, it's always return False, i told you about it, so i can rename this variable )

src/homework06.py
Mantikor: fixed

src/homework06.py
Mantikor: fixed

src/homework06.py
Mantikor: fixed

src/homework06.py
Mantikor: fixed

Mantikor (commented):
bugfixed version is ready

belovmd (commented):
8

PR #106, opened 2019-11-19
Homework6: Aleksandr Strokov

alstrokov:


src/task01.py
```diff
@@ -0,0 +1,27 @@
...
+from persons import Lector
+from persons import Student
+
+
+def main():
+    lector = Lector("Max", 33)
+    stu1 = Student("Nick", 20)
+    stu2 = Student("Ada", 19)
+    group = Group("Q19")
+    group.members.append(stu1)
```
belovmd: usually we use methods to add them (add_member, exclude_member i.e.)

src/materials.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def prepare(self, topic):
+        self.topic = topic
+
+
+class Hometask(Material):
+    tasks = []
+
+    def __init__(self, name):
+        super().__init__(name)
```
belovmd: we don't need super here as parent super would be called if we don't defint init in this class

src/materials.py
```diff
@@ -0,0 +1,43 @@
...
+    @abstractmethod
+    def prepare(self, *param):
+        pass
+
+
+class Lecture(Material):
+    topic = ''
+
+    def __init__(self, name):
+        super().__init__(name)
```
belovmd: redundant init

src/persons.py
```diff
@@ -0,0 +1,61 @@
...
+        self.age = age
+
+    @abstractmethod
+    def greet(self):
+        pass
+
+
+class Student(Person):
+    def __init__(self, name, age):
+        super().__init__(name, age)
```
belovmd: redundant super

belovmd (commented):
please add docstrings to classes and methods
please address these issues

alstrokov:
ready to review

belovmd (commented):
8,5 (maximum score for this task)

PR #107, opened 2019-11-19
Homework4:Nikolai Potovin

adnike:
Sorry, If there is no __init__.py again. Have no idea how to fix it.


src/homework4_.py
```diff
@@ -0,0 +1,64 @@
+# 1 Создайте декоратор, который хранит в себе результаты вызовов функции
+# и время получения результата за время запуска программы.
+import time
+
+global reclist
```
belovmd, line 5: we don't need global here they are already global

src/homework4_.py
```diff
@@ -0,0 +1,64 @@
...
+global call_number
+call_number = 0
+
+
+def eto_moy_decorator(func):
+    """Декорирование функций с разным набором входных параметров"""
+    def obertochka(*args, **kwargs):
+        global call_number
+        call_number += 1
+        print('Вызов', call_number)
```
belovmd, line 16: decorators shouldn't put anything to stdout
it's bad practice

src/homework4_.py
```diff
@@ -0,0 +1,64 @@
...
+        print('Входные параметры', args, kwargs)
+        start_time = time.time()
+        end_time = time.time()
+        result_of_call = ('Итог = ' + str(func(*args, **kwargs)) + '  ' +
+                          'Время вызова = ' + str(end_time - start_time))
+        reclist.append(result_of_call)
```
belovmd, line 22: it's not clear what has been called and measured

src/homework4_.py
```diff
@@ -0,0 +1,64 @@
...
+
+for input_number in range(6):
+    square(input_number)
+    print(reclist[input_number])
+
+
+# 2 Реализовать функцию get_ranges которая получает на вход непустой список
+# неповторяющихся целых чисел, отсортированных по возрастанию,
+# которая этот список “сворачивает”
+def get_ranges(input_list):
```
belovmd, line 42: ['2-3', '6', '8-11', '13', '17-18']
['0-4', '7-8', '10']
['0-4', '7-8', '10']
['-4', '7', '10']
['2-3', '8-9']

is output of this function but should be single string

belovmd (commented):
7

PR #110, opened 2019-11-20
Homework6: Olga Akhmetova

Trylar:


src/homework6/n!.py
```diff
@@ -0,0 +1,48 @@
+"""Algorithm with O(n!) complexity: Travelling salesman problem"""
```
belovmd, line 1: It's too difficult task itself. Surely it has factorial solution (it's pretty obvious) and problem with this task is NP?=P problem. 
During our lectures I've meant regular tasks that are given as home tasks to students :) 🧪

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
+"""Create simulator from real life. This can be booking room in hotel,
+visit to casino, visit to bar. Create 3-4 objects, that can describe
+situation. Objects should contain attributes and methods to simulate some
+use cases. Completed program should print object states, it actions (
+methods) and objects interaction."""
+# flake8: noqa:H238
```
belovmd, line 6: okay :(

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+class FightUnit:
+    def __init__(self, strength=random.randint(1, 20)):
+        self.strength = strength
+
+    def throw_dice(self):
+        return random.randint(1, self.strength + 1)
+
+
+class CrimeFighter(FightUnit):
+    strength = 0
```
belovmd: why do we need this class var?

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class PoliceSquad(CrimeFighter):
+    def fight_gang(self, gang):
+        super().fight_gang(gang)
+        if not self.strength:
+            print("Police squad is destroyed")
+
+
+class Gang(FightUnit):
```
belovmd: I don't understand why it's not Organized Structure but Unit subclass

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+    def is_eliminated(self):
+        self.strength = 0
+        print("Gang was eliminated!")
+
+
+class OrganizedStructure:
+    group_name = ""
+    group_type = FightUnit
```
belovmd: I don't understand why we need these vars here as class variable (remember that instance of this class would be Organized Structure and we can put them to instance variables.

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+    group_size = 3
+    max_group_number = 10
+    reserve = 30
+
+    def __init__(self):
+        setattr(self, self.group_name,
```
belovmd: I completely don't understand this line. Why can't we pass name as parameter to constructor?

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+                [self.group_type(self.group_size) for i in
+                 range(random.randint(1, self.max_group_number))])
+
+    def count_groups(self):
+        return len([group for group in getattr(self, self.group_name) if
```
belovmd: we don't need getattr if we'll refactor this class

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+                    group.strength > 0])
+
+    def get_groups(self):
+        return getattr(self, self.group_name)
```
belovmd: we rarely use getattr and I don't see pro's for this data storage classes

src/homework6/police.py
```diff
@@ -0,0 +1,131 @@
...
+                squad = police.find_squad()
+                if squad:
+                    squad.fight_gang(gang)
+                else:
+                    batman.fight_gang(gang)
+            print("Now we have {} squads and {} gangs and Batman"
+                  .format(police.count_groups(), ocg.count_groups()))
+        print("By the end of {} day we have: {} squads and {} gangs and Batman"
+              .format(day + 1, police.count_groups(), ocg.count_groups()))
+        ocg.restore_groups(), police.restore_groups()
```
belovmd: when should batman be restored? if he wouldn't be lucky he would become 0 power from the first day.

belovmd (commented):
Please put docstrings to classes and methods
Overall it looks as it's too many getattr, setattr and class variables. it seems it should be a bit reworked. 
please take a look at these issues:

src/homework6/n!.py
Trylar: Well, you didn't mentioned that when you said about it, as far as I remember.
In this case you can just shuffle the list until it gets sorted)
I used sorting in my 5 homework

belovmd: > In this case you can just shuffle the list until it gets sorted) 

brilliant 🔥

belovmd (commented):
9 for sorting (Max score for this task is 8.5)

PR #111, opened 2019-11-20
Homework7: Sergey Streltsov

Mantikor:


src/homework07.py
```diff
@@ -0,0 +1,136 @@
...
+    Include else and finally clauses in your try-except block.
+    """
+    try:
+        l = thedict[listname]
+    except KeyError as e:
+        thedict[listname] = []
+        print("Created %s." % listname)
+    else:
+        print("%s already has %d elements." % (listname, len(l)))
+    finally:
```
belovmd, line 122: we can not to use finally here and just put it to function

belovmd (commented):
I really like your tests! good job! 
I've found just one little issue:

src/homework07.py
Mantikor: i don't understood about finally, by condition of the task, we need to use: Include else and finally clauses in your try-except block

belovmd (commented):
9,5

PR #113, opened 2019-11-21
Homework7: Olga Akhmetova

Trylar:


src/homework7/3/tests.py
```diff
@@ -0,0 +1,30 @@
...
+sys.path.insert(1, "../1")
+
+
+class TestRunner(unittest.TestCase):
+    def test_run(self):
+        months = {1: ["January"], 3: ["March"], 4: ["April"], 5: ["May"],
+                  6: ["June"], 7: ["July"], 8: ["August"], 9: ["September"],
+                  10: ["October"], 11: ["November"], 12: ["December"]}
+        output = StringIO()
+        with redirect_stdout(output):
```
belovmd, line 32: wow

src/homework7/3/tests.py
```diff
@@ -0,0 +1,30 @@
...
+from io import StringIO
+import sys
+from task_1_3 import add_to_list_in_dict
+import unittest
+
+sys.path.insert(1, "../1")
+
+
+class TestRunner(unittest.TestCase):
+    def test_run(self):
```
belovmd: usually we use different tests for different cases. just a tip you can use them via ddt library
also they can be just splitted.

src/homework7/2/tests.py
```diff
@@ -0,0 +1,36 @@
+from runner import runner
+import unittest
+
+
+class TestRunner(unittest.TestCase):
+    def test_run(self):
```
belovmd: too many test cases in one test. please split them (you can use ddt lib here to make it more simple)

belovmd (commented):
I really appreciate you stdout redirect
Please take a look at these issues. Usually we don't test a lot of test cases in the same test

belovmd (commented):
9,5

PR #114, opened 2019-11-21
Homework6: Pavel Martysiuk

PavelMartysiuk:


belovmd (commented):
Please provide docstrings to classes and methods

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+        self.salary = randint(1, 15)
+        self.money = 0
+        self.company = False
+        self.company_name = False
+
+
+class Worker(CreateHuman):
+
+    def __init__(self, name, surname):
+        super().create_human(name, surname)
```
belovmd: redundant super

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+              .format(self.name, self.surname, self.salary))
+
+    def work(self, hours):
+        if not self.company:
+            money = int(hours * self.salary)
+            self.money += money
+            print('Salary is {0}. I have {1} dollars now.'
+                  .format(money, self.money))
+        else:
+            print('You work in a company')
```
belovmd: it doesn't mean anything why can't he work somewhere else?

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def dismissal(self):
+        if self.company:
+            self.company.workers.remove(self)
```
belovmd: it's usually works with company class methods not straightly lists.

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class CreateHuman(object):
+    def create_human(self, name, surname):
+        self.name = name
+        self.surname = surname
+        self.salary = randint(1, 15)
+        self.money = 0
+        self.company = False
+        self.company_name = False
```
belovmd: why do we need this name if it would be in company instance?

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+from random import randint
+
+
+class CreateHuman(object):
+    def create_human(self, name, surname):
+        self.name = name
+        self.surname = surname
+        self.salary = randint(1, 15)
+        self.money = 0
+        self.company = False
```
belovmd: usually none

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, company_name, work_cost):
+        self.name = company_name
+        self.work_cost = work_cost
+        self.money = 0
+        self.director = False
+
+    def choose_workers(self, *workers):
+        self.workers = []
+        for worker in workers:
+            if worker.salary < self.work_cost and not \
```
belovmd: if it's too long we should refactor it

src/work.py
```diff
@@ -0,0 +1,146 @@
...
+            worker.money += worker_salary
+            money -= worker_salary
+        if self.director:
+            self.money -= self.director.salary
+            self.director.money += self.director.salary
+        self.money += money
+
+    def show_director(self):
+        if not self.director:
+            return "Company {0} doesn't has a director".format(self.name)
```
belovmd: is it possible in real life?

belovmd (commented):
please put docstrings and address these issues

src/work.py
PavelMartysiuk: We don't need it.

src/work.py
PavelMartysiuk: No

belovmd (commented):
7 (Max score for this task is 8.5)
missed docstrings!

PR #115, opened 2019-11-23
Homework7: Pavel Martysiuk

PavelMartysiuk:


src/exception_practice.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def add_to_list_in_dict(dict_, list_name, element):
+    try:
+        l = dict_[list_name]
+    except KeyError:
+        dict_[list_name] = []
+    else:
+        print('%s already has %d elements.' % (list_name, len(l)))
+    finally:
```
belovmd: we can not use finally here and just put it to method

src/exception_practice.py
```diff
@@ -0,0 +1,49 @@
...
+"""Add a try-except statement to the body of this function which handles a
+possible IndexError, which could occur if the index provided exceeds the
+length of the list. Print an error message if this happens."""
+
+
+def print_list_element(list_, index):
+    try:
+        return list_[index]
+    except IndexError as exc:
+        return str(exc)
```
belovmd: strange to return in first case list and in other string

src/tests_excception_practice.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(print_list_element(self.list_, 0), 1)
+
+    def test_with_index_error(self):
+        self.assertEqual(print_list_element(self.list_, 20),
+                         "list index out of range")
+
+
+class TestDivision(unittest.TestCase):
+    def test_division(self):
+        self.assertEqual(division(10, 2), 5.0)
```
belovmd, line 34: assertRaises can be used here

src/tests_pytasks.py
```diff
@@ -0,0 +1,24 @@
+""" Tests for pytasks.py """
+
+import pytasks
+import unittest
+
+
+class TestPytasks(unittest.TestCase):
+    def test_palindrome(self):
```
belovmd: too many test cases in one test

src/tests_pytasks.py
```diff
@@ -0,0 +1,24 @@
...
+        self.assertEqual(pytasks.palindrome(), True)
+        self.assertEqual(pytasks.palindrome(123), False)
+        self.assertEqual(pytasks.palindrome('sas'), 'Input must be a number')
+
+    def test_get_numbers(self):
```
belovmd: usually we split tests for different test cases i.e. via ddt library

src/tests_pytasks.py
```diff
@@ -0,0 +1,24 @@
...
+        self.assertEqual(pytasks.gen_numbers(), {1: 1, 2: 4, 3: 9})
+        self.assertEqual(pytasks.gen_numbers(5),
+                         {1: 1, 2: 4, 3: 9, 4: 16, 5: 25})
+
+    def test_fizz_buzz(self):
```
belovmd: lack of test cases for this function. 
also please split tests

belovmd (commented):
please take a look at these issues

src/tests_pytasks.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_number(self):
+        self.assertEqual(pytasks.palindrome(123), False)
+
+    def test_str(self):
+        self.assertEqual(pytasks.palindrome('sas'), 'Input must be a number')
+
+
+class TestGetNumbers(unittest.TestCase):
+    def test_default_value(self):
```
belovmd, line 19: usually we test 0, 1 and many (i.e. 2) + default value

src/tests_pytasks.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertEqual(pytasks.gen_numbers(), {1: 1, 2: 4, 3: 9})
+
+    def test_input_value(self):
+        self.assertEqual(pytasks.gen_numbers(5),
+                         {1: 1, 2: 4, 3: 9, 4: 16, 5: 25})
+
+
+class TestFizzBuzz(unittest.TestCase):
+    def test_default_value(self):
```
belovmd, line 28: it's a trick cause fizzbuzz was about printing 100 numbers. 
we should redirect stdout and test not truncated function

src/pytasks.py
```diff
@@ -0,0 +1,31 @@
...
+        else:
+            number_list.append(i)
+    return number_list
+
+
+def gen_numbers(numbers=3):
+    return {digit: digit ** 2 for digit in range(1, numbers + 1)}
+
+
+def palindrome(num=11):
```
belovmd, line 18: why this number should be default?  we should test our numbers, not truncated or reworked ones. it's about all these funcs.

belovmd (commented):
7,5
It shouldn't be so that we change functions for testing. we should test our functions.

PR #118, opened 2019-11-23
Homework6: Aleksandr Levdansky

Coromilly:


warehouse_simulator.py
```diff
@@ -0,0 +1,164 @@
...
+        for material in Warehouse.materials.keys():
+            print('Quantity/Capacity of {mater}: {amount}/{capacity}.'.format(
+                mater=Warehouse.materials[material].material_name,
+                amount=Warehouse.materials[material].material_quantity,
+                capacity=Warehouse.materials[material].material_limit))
+        print('Current number of materials in Warehouse is {}.'.
+              format(len(Warehouse.materials)))
+
+
+class Material(Warehouse):
```
belovmd: it seems strange from abstractions point of view that material is subclass of warehouse

warehouse_simulator.py
```diff
@@ -0,0 +1,164 @@
...
+class Warehouse(object):
+    """This is simulator of Warehouse.
+
+    Starting parameters:
+    material_limit - material limit in m^3
+    money - starting sum of money in dollars
+    cheat - difference between purchase price and sale price
+    extend_cost - cost in dollars of Warehouse extending at 1m^3
+    """
+    material_limit = 5000
```
belovmd: do we really need so many class variables?

warehouse_simulator.py
```diff
@@ -0,0 +1,164 @@
...
+    def info(self):
+        """Shows current info about material."""
+        print('Material - {}: Price: {}, In stock: {}/{}'.format(
+            self.material_name,
+            self.material_price,
+            self.material_quantity,
+            self.material_limit))
+
+
+class Calculator(Material):
```
belovmd: it strange inheritance

belovmd (commented):
please address these issues

warehouse_simulator.py
Coromilly: solved.

warehouse_simulator.py
Coromilly: solved.

warehouse_simulator.py
Coromilly: solved.

warehouse_simulator.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def warehouse_info(self):
+        """Prints materials at Warehouse and their quantity."""
+        for material in Warehouse.materials.keys():
+            print('Quantity of {mater}: {quantity}.'.format(
+                mater=material,
+                quantity=Warehouse.materials[material]))
+        number = 0
+        for value in Warehouse.materials.values():
+            if value != 0:
```
belovmd, line 29: if not value

belovmd (commented):
6,5
Overall it's pretty strange schema
Warehouse should have method to add material to warehouse instance and remove.
Material shouldn't change class (Warehouse) during init. It should just create material inst and possibly have variable that shows warehouse it's kept.
Worker is a person, warehouse is a building. they shouldn't inherit each other
calculator is a tool and it shouldn't be inherited from Warehouse

Coromilly:
> Overall it's pretty strange schema
    I don't think so. 
> Warehouse should have method to add material to warehouse instance and remove.
    Warehouse only stores materials. It's worker's duty to buy/sell materials.
> Material shouldn't change class (Warehouse) during init. It should just create material inst and possibly have variable that shows warehouse it's kept.
    Material can only change the dicts 'materials' and 'price' in Warehouse class. It is needed for getting simple info about Warehouse.
> Worker is a person, warehouse is a building. they shouldn't inherit each other
    I agree.
> calculator is a tool and it shouldn't be inherited from Warehouse
    I wanted to show the inheritance property of classes.

PR #120, opened 2019-11-24
Homework7: Aleksandr Levdansky

Coromilly:


task1_3.py
```diff
@@ -0,0 +1,30 @@
...
+try-except block.
+"""
+
+
+def add_to_list_in_dict(thedict, listname, element):
+    try:
+        try:
+            lst = thedict[listname]
+        except TypeError:
+            raise TypeError("Input correct arguments.")
```
belovmd: you can print error and just raise without anything

task1_3.py
```diff
@@ -0,0 +1,30 @@
...
+        print("%s already has %d elements." % (listname, len(lst)))
+    except KeyError:
+        thedict[listname] = []
```
belovmd: ...] = [element]?

task3.py
```diff
@@ -0,0 +1,21 @@
+"""Cover function from 1.3 with tests."""
+from task1_3 import add_to_list_in_dict
+import unittest
+
+
+class FuncTest(unittest.TestCase):
+    def setUp(self):
```
belovmd: it's strange we put all these constants to setup but use in different tests different values. 
usually test is setup, action, assert. your one is just equal

task2.py
```diff
@@ -0,0 +1,32 @@
...
+
+def count_characters(s):
+    """Function count and return the numbers of each character in a string."""
+    dct = {}
+    for symb in s:
+        dct[symb] = dct.get(symb, 0) + 1
+    return dct
+
+
+class TestCountCharacters(unittest.TestCase):
```
belovmd, line 13: why in count characters test collection are upper lower etc?

task2.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_normal(self):
```
belovmd, line 14: we need more tests for this function

belovmd (commented):
please address these issues

task1_3.py
Coromilly: solved.

task1_3.py
Coromilly: solved.

task3.py
Coromilly: solved.

task2.py
Coromilly: solved.

task2.py
Coromilly: solved.

task3.py
```diff
@@ -0,0 +1,19 @@
+"""Cover function from 1.3 with tests."""
+from task1_3 import add_to_list_in_dict
+import unittest
+
+
+class FuncTest(unittest.TestCase):
```
belovmd, line 6: why here are only string keys?

task2.py
```diff
@@ -0,0 +1,37 @@
...
+        res = count_characters('coromilly')
+        self.assertEqual(
+            res, {'c': 1, 'o': 2, 'r': 1, 'm': 1, 'i': 1, 'l': 2, 'y': 1})
+
+    def test_default_value(self):
```
belovmd, line 19: it has strange default value. i.e. we can call it with empty args and check assert raises

belovmd (commented):
8,5

PR #124, opened 2019-11-24
Homework7: Anton Koliada

JustgoodDeal:


Homework7/Task1.py
```diff
@@ -0,0 +1,52 @@
...
+
+def add_to_list_in_dict(thedict, listname, element):
+    try:
+        ll = thedict[listname]
+    except KeyError:
+        thedict[listname] = []
+        print("Created %s." % listname)
+    else:
+        print("%s already has %d elements." % (listname, len(ll)))
+    finally:
```
belovmd: we don't need finally here

Homework7/Task3.py
```diff
@@ -0,0 +1,48 @@
+""" Cover function from 1.3 with tests. """
+
+import unittest
+
+
+def add_to_list_in_dict(thedict, listname, element):
```
belovmd, line 6: duplicated code please use your solution

belovmd (commented):
please address these issues

belovmd (commented):
8,5
there are two practically similar functions here (it was mentioned in last review). 
we should use import here.

PR #128, opened 2019-11-24
Homework7 Nikolai Potovin

adnike:
Nikolai Potovin


src/Homework7.py
```diff
@@ -0,0 +1,92 @@
...
+
+def add_to_list_in_dict(thedict, listname, element):
+    try:
+        l = thedict[listname]
+    except KeyError:
+        thedict[listname] = []
+        print("Created %s." % listname)
+    else:
+        print("%s already has %d elements." % (listname, len(l)))
+    finally:
```
belovmd, line 42: we don't need finally here

src/Homework7.py
```diff
@@ -0,0 +1,92 @@
...
+
+class MyTestFunction(unittest.TestCase):
+
+    def test_add_without_list(self):
+        legends_1 = {'Beatles': ['John_lennon'],
+                     'Rolling_Stones': ['Mick_Jagger'],
+                     'Deep_Purple': ['Ian_Gillan'],
+                     'Led_Zeppelin': ['Robert_Plant', 'Jimmy_Page']}
+        add_to_list_in_dict(legends_1, 'Queen', 'Freddie_Mercury')
+        legends_2 = {'Beatles': ['John_lennon'],
```
belovmd, line 71: we can name it expected_dict

src/Homework7.py
```diff
@@ -0,0 +1,92 @@
...
+# http://nose.readthedocs.io/en/latest/plugins/cover.html
+
+# Sorry, have problems with installing nosetest on windows at this moment,
+# I will add this task later
+# 3 Cover function from 1.3 with tests
+
+
+class MyTestFunction(unittest.TestCase):
+
+    def test_add_without_list(self):
```
belovmd, line 65: we can add test with empty started dict

belovmd (commented):
Overall it looks good, please take a look at these issues:

PR #131, opened 2019-11-28
Homework6: Pavel Ralko

par216:


Hw6_1.py
```diff
@@ -0,0 +1,36 @@
+class Person(object):
+    def __init__(self, name, age, money):
+        self.name = name
+        self.age = age
+        self.money = money
+        print("Information about the person:", name, age, money)
+
+
+class Dealer(Person):
+    def start_game(self):
```
belovmd, line 10: it seems here should be as a parameters gamers who will participate in game. 
here also should be generated game results.

Hw6_1.py
```diff
@@ -0,0 +1,36 @@
...
+        print("The game began")
+
+
+class Gamer(Person):
```
belovmd, line 14: he should have not win and loose methods but get money and spend_money ones. 
who has wins and amount of money should be desided by a dealer

belovmd (commented):
6 (maximum score for this task is a 8,5)
1. Overall it's too high level models with misses in classes structure
2. missed docstrings for classes and methods

PR #132, opened 2019-11-29
Homework7: Pavel Ralko

par216:


test.py
```diff
@@ -0,0 +1,11 @@
+from palindrome import palindrome
+import unittest
+
+
+class TestMetods(unittest.TestCase):
```
belovmd, line 5: call without args, n<0 hasn't been tested

test777.py
```diff
@@ -0,0 +1,18 @@
+from Homework7 import add_to_list_in_dict
+from Homework7 import dict1
+import unittest
+
+
+class TestFunction(unittest.TestCase):
+
+    def test1(self):
+        add_to_list_in_dict(dict1, 2, 3)
```
belovmd, line 9: we shouldn't use side mutable variables in tests - they would affect code.

belovmd (commented):
6
very few tests, many test cases were missed, tests change global vars in code

PR #133, opened 2019-11-29
Homework7 Nikolai Potovin

adnike:
changes after review


src/Homework7.py
```diff
@@ -0,0 +1,124 @@
...
+    while input_number != 0:
+        ostatok = input_number % 10
+        reversd_num = reversd_num * 10 + ostatok
+        input_number = input_number // 10
+    if original_num != reversd_num:
+        return False
+    return True
+
+
+class MyTestFunction(unittest.TestCase):
```
belovmd, line 80: we can also check palindrome without string

belovmd (commented):
9

PR #136, opened 2019-11-29
Homework5: Andrey Kapitonov

akapitonov1999:


runner.py
```diff
@@ -0,0 +1,25 @@
+import pytasks
+
+
+def runner(a='', b='', c='', d=''):
```
belovmd, line 4: we can use just getattr.  it would be much more flexible

belovmd (commented):
In fact here it is weird runner but just excellent histograms. 
9

PR #143, opened 2019-12-04
Homework6: Andrey Kapitonov

akapitonov1999:


Homework6.py
```diff
@@ -0,0 +1,157 @@
...
+
+        bacteria_count - contains a list, where each bacteria
+        is an integer [1], so we'll be able to manipulate
+        with each bacteria in future
+
+        """
+
+        self.species = species
+        self.risk = risk
+        self.bacteria_count = list(1 for x in range(bacteria_count))
```
belovmd, line 37: overall too many range len etc. more of that do we really need list [1,1,1...] for bacterias? could we just use int?

Homework6.py
```diff
@@ -0,0 +1,157 @@
...
+    class Bacteria - to describe bacterial cells
+    class Medium - to describe environment and the experiment
+    class Scientist - to describe and visualize the results
+    and to describe the scientist
+    function go_on - to start and run the experiment properly'''
+
+
+class Bacteria(object):
+
+    exp_data = list()
```
belovmd, line 20: it's not good idea to keep all bacterias list and count in class variable for bacteria. 
we should create bacterias class for this.
overall, class variables are rarely used option. Here it's not needed.

belovmd (commented):
8 (maximum score for this task is 8,5)
overall it's implemented not in a best way (strange solution with class variables, missing vitro class where we can store all bacteries). 
But idea is unusual and have drawings ^) 

Overall< when I've read docstring I've suspected smth like this 
https://studfile.net/preview/4283559/page:2/
there are updated math models for bad conditions and so on (it's first link I've found). Population models are classic models for differential equations. But it seems they are used only in books about equations.

PR #144, opened 2019-12-04
 Homework4: Maksim Bleo

Meolax:


Meolax:
flake8_error

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
...
+    def __call__(self, value):
+        """."""
+        changed = +(value - self.prev) != 1
+        self.prev = value
+        if changed:
+            self.gr_id = value
+        return self.gr_id
+
+
+def get_ranges(arr):
```
belovmd: it should be this

src/task_4_1_decorator.py
```diff
@@ -0,0 +1,47 @@
...
+def func_logs(func):
+    """Save func logs .
+
+    Decorator which save logs for last sequence call of one function.
+    """
+    logs = []
+
+    def wrapper(*args, **kwargs):
+        func_result = func(*args, **kwargs)
+        logs.append({'func_name': func.__name__,
```
belovmd, line 21: we can access this via func_name.logs that's why we don't need name here

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Реализовать функцию get_ranges которая получает на вход непустой
+    список неповторяющихся целых чисел, отсортированных по возрастанию,
+    которая этот список “сворачивает”
+    get_ranges([0, 1, 2, 3, 4, 7, 8, 10]) // "0-4,7-8,10"
+    get_ranges([4,7,10]) // "4,7,10"
+    get_ranges([2, 3, 8, 9]) // "2-3,8-9"
+    """
+    ranges = ''
+    for gr_id, values in groupby(arr, OneFromLast(arr[0])):
```
belovmd, line 34: it looks overhead here. It doesn't became easier to read or lesser because of using lib so why do we use it?

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
...
+from itertools import groupby
+
+
+class OneFromLast():
+    """."""
+
+    def __init__(self, first):
+        """."""
+        self.prev = first
+        self.gr_id = first
```
belovmd, line 12: what does gr_id means?

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
+"""Task_4_2."""
+
+from itertools import groupby
+
+
+class OneFromLast():
+    """."""
```
belovmd, line 7: I don't understand this docstring

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self, first):
+        """."""
```
belovmd, line 10: same issue

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
...
+    get_ranges([4,7,10]) // "4,7,10"
+    get_ranges([2, 3, 8, 9]) // "2-3,8-9"
+    """
+    ranges = ''
+    for gr_id, values in groupby(arr, OneFromLast(arr[0])):
+        group_list = list(values)
+        if len(group_list) == 1:
+            ranges += wrap_num(group_list[0]) + ','
+        else:
+            ranges += wrap_num(group_list[0]) + '-' + \
```
belovmd, line 39: do we really need \\ here? it can be refactored

src/task_4_2_get_ranges.py
```diff
@@ -0,0 +1,55 @@
+"""Task_4_2."""
```
belovmd, line 1: Overall it looks as overhead
it's task for technical interview. 
it seems practically impossible to implement such solution in 10-15 minutes and it in fact doesn't have any advantages.

belovmd (commented):
Overall it looks like 7,5-8
but:
 - 12 days overdue  -1.5
- failed CI -1 (it's not too difficult to put object to superclasses and we have restriction that tests should pass)
final score 5.5 (8-1,5-1)

PR #145, opened 2019-12-04
Homework6: Anastasiya Infarovich

AnInf:


src/task6/oop.py
```diff
@@ -0,0 +1,101 @@
+"""Simulator from real life medical drug selling.
+
+Simulated objects contain attributes and methods to simulate some use cases.
+Program prints objects' states, methods and interaction.
+"""
+
+import types
+
+
+class DecoMeta(type):
```
belovmd, line 10: wow, metaclasses!!!

belovmd (commented):
Maximum score for this task according task description is 8.5
Overal it's smth like 7.5 because of classes and methods doesn't have docstrings and it's pretty strange work with attributes that should be private (money)
But it's first time I see metaclasses and decorators in this class (no matters we doesn't need metaclasses here (they has rare usage)
Let it be final  8.5 - maximum score for this task

PR #148, opened 2019-12-05
Homework7:Anastasiya Infarovich

AnInf:


src/task7/2_test.py
```diff
@@ -0,0 +1,22 @@
+from digits_multiplication import checkio
+import unittest
+
+
+class CheckioTest(unittest.TestCase):
+
+    def test_valid(self):
```
belovmd, line 7: we can use ddt with value, expected value and make 3 tests from this one for example

src/task7/3_test.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class AddToListInDict(unittest.TestCase):
+
+    @unittest.mock.patch('sys.stdout', new_callable=io.StringIO)
+    def assert_stdout(self, n, expected_output, mock_stdout):
+        add_to_list_in_dict(*n)
+        self.assertEqual(mock_stdout.getvalue(), expected_output)
+
+    def test_barBaz(self):
```
belovmd, line 15: not pythonic naming we_use_snake_case

belovmd (commented):
9

PR #149, opened 2019-12-08
Homework7: Andrey Kapitonov

akapitonov1999:


Homework6.py
```diff
@@ -0,0 +1,171 @@
+import matplotlib.pyplot as plt
+import random
+import unittest
+
+plt.style.use('ggplot')
+
+# task 2
+# Actually, I'm not sure about this task, because nosetests shows 95% coverage
+# for this file, but 1% TOTAL. Seems that nosetests wants me to cover with
+# tests matplotlib module.
```
belovmd, line 10: you should measure it smth like nosetest /file --with-coverage --cover-erase --cover-package=./ and it would be ok (it wouldn't count external modules)

Homework6.py
```diff
@@ -0,0 +1,171 @@
...
+        bact.internal_death()
+        bact.bacterias_dividing()
+        med.experiment_stage1(bact.bacteria_count)
+    med.experiment_stage2(bact.bacteria_count)
+    med.experiment_stage3(bact.bacteria_count)
+    sci.results()
+
+
+class MyTest(unittest.TestCase):
+    def test(self):
```
belovmd, line 164: bad test naming - it should describe a bit what we test)

homework7.py
```diff
@@ -0,0 +1,50 @@
...
+        print("%s already has %d elements." % (listname, len(curr_list)))
+    finally:
+        thedict[listname].append(element)
+        print("Added %s to %s." % (element, listname))
+    return thedict
+
+
+# task 3
+class MyTest(unittest.TestCase):
+    def test(self):
```
belovmd, line 43: bad test naming

Homework6.py
```diff
@@ -0,0 +1,171 @@
...
+        self.assertEqual(go_on('Aspergilla', 0.3, 4, 'Chocolate agar',
```
belovmd, line 165: it doesn't have what to assert to

belovmd (commented):
Usually we write atomic little tests - they cover one method i.e. or one user action. 
In fact your test tests practically all functionality. It will help sometimes to figure out that smth went wrong but wouldn't help us to find what exactly
7

